PR #69, opened 2026-04-06
Fix multi-repo support

lfrancke:
## What does this PR do?

The tool would bail out too early so far. It would detect that there is no git directory and then bail instead of checking for subdirectories with git directories.

I also had to skip the `git check-ignore` check when there is no `.git` directory.

I tested it locally and it seems to work fine for me!
I tested it in a normal git repo, in a parent directory with multiple sub-repos and in a directory with no repo at all.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

- [ ] I've updated documentation if needed

### DIFF
--- a/cmd/hooks.go
+++ b/cmd/hooks.go
@@ -247,18 +247,20 @@ func RunHook(hookName, root string) error {
 
 // hookSessionStart shows project structure, starts daemon, and shows hub warnings
 func hookSessionStart(root string) error {
-	// Guard: require git repo
+	// Handle meta-repo: parent directory containing child git repos.
+	// Check this before the git guard so that non-repo parent directories
+	// still get multi-repo context.
+	childRepos := findChildRepos(root)
+	if len(childRepos) > 1 {
+		return hookSessionStartMultiRepo(root, childRepos)
+	}
+
+	// Guard: require git repo for single-repo analysis
 	gitDir := filepath.Join(root, ".git")
 	if _, err := os.Stat(gitDir); os.IsNotExist(err) {
 		fmt.Println("📍 Not a git repository - skipping project context")
 		fmt.Println("   (codemap hooks work best in git repos)")
 		return nil
-	}
-
-	// Handle meta-repo: parent git repo containing child git repos
-	childRepos := findChildRepos(root)
-	if len(childRepos) > 1 {
-		return hookSessionStartMultiRepo(root, childRepos)
 	}
 
 	// Check for previous session context before starting new daemon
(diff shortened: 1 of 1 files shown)


cmd/hooks.go
```diff
@@ -247,20 +247,22 @@ func RunHook(hookName, root string) error {
 
 // hookSessionStart shows project structure, starts daemon, and shows hub warnings
 func hookSessionStart(root string) error {
-	// Guard: require git repo
+	// Handle meta-repo: parent directory containing child git repos.
```
lfrancke, line 250: This changes no code: It only swaps the multiRepo check with the "git guard"

cmd/hooks.go
```diff
@@ -1453,15 +1455,18 @@ func findChildRepos(root string) []string {
...
-	cmd := exec.Command("git", args...)
-	cmd.Dir = root
-	out, _ := cmd.Output()
+	// Use git check-ignore to filter out ignored directories.
+	// Skip filtering if the root is not a git repo (e.g. a parent directory).
 	ignored := make(map[string]bool)
-	for _, line := range strings.Split(strings.TrimSpace(string(out)), "\n") {
-		if line != "" {
-			ignored[line] = true
+	if _, err := os.Stat(filepath.Join(root, ".git")); err == nil {
```
lfrancke: This just skips `git check-ignore` if there is no `.git` directory

JordanCoin:
Thanks for taking this on, and for spelling out the cases you tested in the PR description (`normal git repo`, `parent dir with multiple sub-repos`, and `directory with no repo`). The direction looks right to me: moving child-repo detection ahead of the top-level git guard is what unlocks the non-git multi-repo parent case.

The one case I think is still missing is: a non-git parent with multiple child repos where the parent `.gitignore` excludes one of those child repo directories. In that setup, this branch will still include the ignored child, because `git check-ignore` is skipped entirely when the parent has no `.git`.

Since `findChildRepos` is documented as honoring the parent ignore list, could you preserve that behavior in the new path and add a regression test for it? I think that would get this over the line.

lfrancke:
Thanks for looking at it. Phew... I'm not sure what to do to be honest.

> a non-git parent with multiple child repos where the parent .gitignore 

I'd say if we're in a non-git parent then a `.gitignore` is meaningless and we can ignore it?
I believe this is already the current behavior because you ignore the output (it could be an error) of `git check-ignore`: `out, _ := cmd.Output()` and then just try to split the return.

So, I just reverted my change because the original code already handles non-git parrent correctly. I missed that bit to be honest and the explicit `.git` check I added was redundant. Unless you _want_ the behavior to change?

lfrancke:
@JordanCoin one quick ping

PR #75, opened 2026-07-02
Make `watch stop` work safely on Windows (#74)

### LINKED ISSUE #74: Make `watch stop` work safely on Windows (PID-ownership verification)
Follow-up from #73 / #72.

#73 fixes `watch status` on Windows (via a platform-specific `processAlive`). It intentionally leaves `watch stop` non-destructive on Windows (SIGTERM is unsupported, so it returns an error without killing).

Making Windows `stop` actually terminate the daemon requires guarding against PID reuse: a stale `.codemap/watch.pid` may point to a PID the OS has reassigned to an unrelated process, and blindly `Kill`-ing it would terminate the wrong process (P2 raised by Codex on #73).

### Work
- Implement `processCommandLine` on Windows (currently a stub) — e.g. via `Get-CimInstance Win32_Process` / `QueryFullProcessImageName` — so `IsOwnedDaemon` can verify the PID is a codemap `watch daemon <root>` for this repo.
- Gate the Windows stop/kill path on that ownership check; refuse (with a clear message) when ownership can't be established.
- Consider applying the same ownership guard to the Unix stop path for consistency.

### Acceptance
- `codemap watch stop` terminates a real daemon on Windows.
- A stale `watch.pid` whose PID was reused does **not** get killed.

JordanCoin:
Closes #74. Follow-up to #73 (which fixed `watch status` on Windows).

## What
Makes `codemap watch stop` actually terminate the daemon on Windows, while guarding against the PID-reuse risk Codex flagged on #73.

- **`processCommandLine` on Windows** (was a stub): now queries the full command line via `Get-CimInstance Win32_Process` — the Windows analog of the Unix `ps` call — so `IsOwnedDaemon` can confirm a PID is this repo's `watch daemon <root>`.
- **Platform-specific `terminateDaemon`**:
  - **Unix**: unchanged — ungated `SIGTERM` (preserves long-standing behavior + existing tests).
  - **Windows**: verifies `IsOwnedDaemon` first, then `Kill`; returns `ErrForeignDaemonPID` when ownership can't be confirmed (refuse-when-uncertain).
- **`Stop`**: on a stale/foreign PID it clears the bogus `watch.pid` (so status stops reporting it) but never kills the unrelated process. The `stop` command reports `Watch daemon not running (cleared stale PID file)`.

## Why Windows-only gating
Killing a reused PID is destructive; on Windows we refuse when uncertain. Unix keeps its historical ungated `SIGTERM` (a reused-PID `SIGTERM` is far less catastrophic, and gating it would risk refusing to stop a real daemon in environments where process introspection is restricted).

## ⚠️ Testing limitation
The Windows runtime path (CIM command-line lookup + ownership kill) **cannot be exercised in this environment or in CI** — the Windows CI job only *builds*, it doesn't run tests. Verified here:
- `GOOS=windows go build` + `go vet` clean.
- Host (Unix) tests green, including a new command-layer test for the foreign-PID branch.

@ScottRosenberg2 — if you're able, verification on real Windows would be valuable:
```powershell
codemap watch start
codemap watch stop     # should stop the daemon
codemap watch status   # should report not running
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -597,6 +597,10 @@ func runWatchSubcommand(subCmd, root string) {
 			return
 		}
 		if err := stopWatchDaemon(absRoot); err != nil {
+			if errors.Is(err, watch.ErrForeignDaemonPID) {
+				fmt.Println("Watch daemon not running (cleared stale PID file)")
+				return
+			}
 			fmt.Fprintf(os.Stderr, "Error stopping daemon: %v\n", err)
 			os.Exit(1)
 		}
--- a/main_more_test.go
+++ b/main_more_test.go
@@ -758,3 +758,24 @@ func TestSubcommandDispatchViaBinary(t *testing.T) {
 		}
 	})
 }
+
+func TestRunWatchSubcommandStopForeignPID(t *testing.T) {
+	root := t.TempDir()
+
+	origRunning := watchIsRunning
+	origStop := stopWatchDaemon
+	defer func() {
+		watchIsRunning = origRunning
+		stopWatchDaemon = origStop
+	}()
+	watchIsRunning = func(string) bool { return true }
+	stopWatchDaemon = func(string) error { return watch.ErrForeignDaemonPID }
+
+	stdout, _ := captureMainStreams(t, func() { runWatchSubcommand("stop", root) })
+	if !strings.Contains(stdout, "cleared stale PID file") {
+		t.Fatalf("expected stale-PID message on foreign PID, got:\n%s", stdout)
+	}
+	if strings.Contains(stdout, "Watch daemon stopped") {
+		t.Fatalf("should not claim the daemon was stopped for a foreign PID:\n%s", stdout)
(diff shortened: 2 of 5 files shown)


JordanCoin:
Thanks @Copilot — both addressed in the latest commit:

1. **TOCTOU**: `terminateDaemon` now classifies ownership against `proc.Pid` (the exact process it's about to kill) instead of re-reading `watch.pid`, so a concurrent start/stop can't cause it to validate one PID and kill another.
2. **Wording / refuse-when-uncertain**: ownership is now three-state (owned / foreign / unknown). `ErrForeignDaemonPID` is only returned when the command line was read and genuinely didn't match (and `Stop` clears the stale pid file then). When the command line can't be read, it returns the new `ErrDaemonOwnershipUnknown` and **keeps** the pid file — so a real daemon is never orphaned and no unrelated process is killed.

PR #77, opened 2026-07-12
feat(config): surface matches hidden by extension filters

reneleonhardt:
## What does this PR do?

- Make MCP file searches respect the project's `only` and `exclude` filters.
- Apply those filters consistently to context, structure, project-statistics, dependency, and file-graph scans.
- When `only` hides real matches, report them and explain how to include or suppress their extensions.
- Enable extension hints by default with `guidance.missing_extension_hints`; set it to `false` to disable them.
- Allow selected hints to be suppressed with `guidance.ignored_extensions`.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

## Additional notes

Repository languages evolve after initial setup. Actionable misses prevent agents from repeatedly failing or silently falling back to raw filesystem searches, without rescanning or changing config automatically.

Focused MCP, context, dependency, and file-graph tests pass with the full race
suite, formatting, module verification, vet, staticcheck, and a built-binary
parity probe. The probe confirmed the CLI context, MCP structure, and MCP project
statistics all report the same configured file count.

Reviewer setup:

```bash
set -euo pipefail

ROOT=$PWD
BRANCH=feat/config-only-fallback
REVIEW_ROOT=$(mktemp -d "${TMPDIR:-/tmp}/codemap-config-fallback.XXXXXX")
WT="$REVIEW_ROOT/worktree"
DEBUG_BIN="$REVIEW_ROOT/codemap-debug"
COVERAGE_OUT="$REVIEW_ROOT/coverage.out"

git worktree add --detach "$WT" "$BRANCH"

cleanup() {
  cd "$ROOT"
  git worktree remove --force "$WT"
  rm -rf "$REVIEW_ROOT"
}
trap cleanup EXIT

cd "$WT"
go fmt ./...
git diff --exit-code
go mod verify
go vet ./...
staticcheck ./...
go test -race -coverprofile="$COVERAGE_OUT" ./...
go build -gcflags='all=-N -l' -o "$DEBUG_BIN" .
"$DEBUG_BIN" .
cleanup
trap - EXIT
```

Developed with carefully directed, manually reviewed AI assistance. `Allow edits by maintainers` is enabled, so maintainers are welcome to fix wording or make minor cleanup directly; broader review feedback will be incorporated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,10 @@ Example `.codemap/config.json`:
 {
   "only": ["rs", "sh", "sql", "toml", "yml"],
   "exclude": ["docs/reference", "docs/research"],
+  "guidance": {
+    "missing_extension_hints": true,
+    "ignored_extensions": []
+  },
   "depth": 4,
   "mode": "auto",
   "budgets": {
@@ -330,6 +334,8 @@ Example `.codemap/config.json`:
   }
 }
 ```
+
+When an MCP file search has no visible results but finds real matches hidden by `only`, Codemap reports the matching paths and suggests which extensions to include. These hints still respect `exclude` and never modify project config. Set `guidance.missing_extension_hints` to `false` to disable all hints, or list extensions in `guidance.ignored_extensions` to suppress only those suggestions.
 
 All fields are optional. CLI flags always override config values.
 Hook-specific policy fields are optional and bounded by safe defaults.
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -307,7 +307,7 @@ func detectLanguagesFromFiles(root string) map[string]bool {
 
 	// Include subdirectory source files. Reuse the scan result for countSourceFiles too.
 	gitCache := scanner.NewGitIgnoreCache(root)
-	if files, err := scanner.ScanFiles(root, gitCache, nil, nil); err == nil {
+	if files, err := scanner.ScanConfiguredFiles(root, gitCache); err == nil {
 		for _, f := range files {
 			addLang(scanner.DetectLanguage(f.Path))
 		}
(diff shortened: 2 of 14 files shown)


JordanCoin (approved):
Reviewed the logic and tests carefully — the core filter/hint design is sound. Hinted files must fail `only` but pass `exclude`, so explicitly excluded paths are never leaked; the hint is bounded (5 paths + "and N more"); and I especially appreciated `TestHandleFindFileExplainsOnlyFilteredMatches` asserting config.json is byte-identical after the call. Default-on for `guidance.missing_extension_hints` is reasonable since it only fires on zero-result searches and never mutates config. Full suite passes with `-race`, and the tests genuinely encode the new behavior (they fail against main's implementation).

Approving, with a few comments — mostly things I want on the record:

1. **Scope is broader than the title suggests.** Beyond `find_file` hints, this makes `get_structure`, `get_diff`, project stats, `BuildFileGraph`/`ScanForDeps`, the CLI context envelope, hooks, and blast-radius output all apply config `only`/`exclude`. I think that's the *right* behavior, but it's a real change on filtered repos, and it creates a gap: `ScanConfiguredFiles`/`filterConfiguredAnalyses` do a hidden `config.Load` inside the scanner, so `--deps`/`--importers` now apply config filters with no way for CLI `--only`/`--exclude` to override — at odds with the documented "CLI flags always override config values." A follow-up plumbing filters as parameters (or documenting the exception) would be welcome.
2. **Hint wording** (`mcp/find_guidance.go`): the response says "Tell your agent: 'include suggestions for proto…'" — but the reader of that text *is* the agent, and the quoted phrase doesn't map to a concrete action. Suggest actionable wording like: "To surface these, add \"proto\" to `only` in .codemap/config.json".
3. Minor: analyses get filtered twice (`ScanForDeps` and `BuildFileGraphFromAnalyses`), with three `config.Load` disk reads per graph build — worth consolidating sometime.

None of these block the merge. Thanks for the thorough test coverage — the entrypoint parity test is a nice touch.

PR #86, opened 2026-07-28
fix(scanner): Resolve imports conservatively

reneleonhardt:
## What does this PR do?

Limit fuzzy resolution to compatible language families. Resolve Go imports only to indexed, non-test files in the current module; keep bare JS/TS packages external unless a configured alias proves an included local target.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other

## Verification

- `go test ./scanner -run '^(TestBuildFileGraphImportResolutionBoundaries|TestFilteredImportResolutionTargets)$' -count=1`
- `go test ./scanner -count=1`

Developed with carefully directed, manually reviewed AI assistance.

Co-Authored-By: GPT-5.6 Sol <codex@openai.com>

PR #87, opened 2026-07-28
fix(scanner): Resolve Rust workspace imports

reneleonhardt:
## What does this PR do?

- Resolve `mod` declarations only within their legal Rust module tree.
- Resolve fully-qualified crate calls from Cargo workspace membership and package names.
- Report partial coverage when dynamic routes cannot be inferred.
- Verify a known cross-layer dependency after Codemap tuning.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [x] Documentation
- [ ] Other

## Verification

- `go test ./scanner -run '^TestRustWorkspace' -count=1`

### DIFF
--- a/blast_radius.go
+++ b/blast_radius.go
@@ -81,6 +81,8 @@ type blastRadiusImporters struct {
 	ImportersTotal  int      `json:"importers_total"`
 	ImportsTotal    int      `json:"imports_total"`
 	HubImportsTotal int      `json:"hub_imports_total"`
+	CoverageStatus  string   `json:"coverage_status,omitempty"`
+	CoverageNotes   []string `json:"coverage_notes,omitempty"`
 }
 
 type blastRadiusHighest struct {
@@ -612,6 +614,8 @@ func capBlastRadiusImportersReport(report scanner.ImportersReport, max int) blas
 		ImportersTotal:  len(report.Importers),
 		ImportsTotal:    len(report.Imports),
 		HubImportsTotal: len(report.HubImports),
+		CoverageStatus:  report.CoverageStatus,
+		CoverageNotes:   append([]string(nil), report.CoverageNotes...),
 	}
 }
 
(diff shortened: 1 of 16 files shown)


JordanCoin:
Heads up for sequencing: #84 and #85 just landed on main, and #86 will need a merge-up/rebase against them (details in the comment there) before this one. Expect `scanner/filegraph.go` and the scanner tests to have moved underneath this branch — probably worth rebasing this after #86 settles rather than before. No review concerns from a first read of the Rust module-tree/Cargo-membership approach; will do a full review once it's rebased.

reneleonhardt:
Rebased onto current `main` without problems.
The reconciliation retains main’s scanner graph contracts alongside the Rust module-tree and Cargo-workspace resolution.
The focused Rust and scanner truth tests, full race-and-coverage suite, vet, staticcheck, build, and `--deps` smoke test are green.
Ready for full review.

JordanCoin:
Ran the merge-up against post-#86 main locally before merging this. Findings:

**Mechanical (fine):** this branch is based on a pre-final iteration of #86, so the squash of #86 produces one conflict hunk in `BuildFileGraphFromFilteredAnalyses` — your `a.Language == "rust"` branch composes cleanly with #86-final's updated `isGoPkg` line (`DetectLanguage` + `isLocalGoImport`). Everything else auto-merges, and the full `-race` suite passes except:

**Semantic (your call):** #86-final's `TestBuildFileGraphImportResolutionBoundaries` expects `crates/path.rs` ← `crates/mod.rs` via `crate::path`, but that fixture has **no Cargo.toml** — so `resolveRustReferences` with an empty workspace index resolves nothing and the assertion fails. As far as I can tell your resolver is behaving exactly as designed (no workspace membership → fail closed), and it's the #86 fixture that's now under-specified: it needs either a minimal `Cargo.toml` to make the edge legal, or an updated expectation (empty + partial-coverage note). Since both tests are yours, I'll leave which way to take it to you rather than pick in a merge commit.

Happy to push the merge-up with whichever fixture resolution you prefer, or take your rebase — either works. Everything else in the PR reviewed well: the Cargo membership/exclude globbing, crate-name normalization, and the honest `GraphCoverage` partial signal are all solid.

reneleonhardt:
Rebased onto PR 86 / main and addressed the fixture finding:
the test now declares the minimal Cargo crate and supplies the structured rust-path reference emitted by the analyzer.
Resolver behavior remains fail-closed.

PR #95, opened 2026-08-02
fix(watch): Track configured file counts separately

reneleonhardt:
## What does this PR do?

Tracks configured files separately from the full watch inventory, including live config and `.gitignore` changes.
Session hooks therefore size agentic coding context from active sandbox filters without changing the legacy `file_count` field or publishing stale dependency state.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

## Additional notes

Verification: `go test ./watch ./cmd -count=1`.

### DIFF
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -193,6 +193,7 @@ func buildProjectContext(root string, info *hubInfo) ProjectContext {
 	ctx := ProjectContext{
 		Root: root,
 	}
+	configuredFileCountKnown := false
 
 	// Get branch
 	if branch, ok := gitCurrentBranch(root); ok {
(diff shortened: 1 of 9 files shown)


JordanCoin:
Reviewed — this is careful work. Using `*int` + `ConfiguredCount() (int, bool)` so a stale state file is distinguishable from a genuine zero is exactly right, and separating `FileCount` (all tracked) from `ConfiguredFileCount` (filter-included) fixes a real conflation in the large-repo thresholds.

One functional finding:

**`refreshConfiguredFiles` permanently drops the dependency graph until the daemon restarts.** It nils `FileGraph`, clears `DepCtx`, and sets `HasDeps = false`, with the comment "rebuild lazily on restart" — but there's no lazy rebuild anywhere; nothing recomputes it. So editing `.codemap/config.json` (or touching any `.gitignore`, since `filterControlEvent` matches by basename anywhere in the tree) silently kills live hub/importer intelligence for the rest of the daemon's life. Since you're already re-walking in that function, calling `computeDeps()` behind the same `shouldComputeDependencyGraph` guard would keep the invariant you're protecting without the cliff.

Minor: `isConfiguredFile` calls `config.Load(d.root)` per file event, so a bulk change (branch switch, install) re-reads and re-parses `config.json` once per file. Worth hoisting alongside the refresh.

Nice catch excluding `.codemap/` from `ScanConfiguredFiles` now that the daemon watches that directory — that would otherwise have inflated the count via its own state writes.

PR #105, opened 2026-08-04
fix(scanner): Consolidate scanner, MCP, and analysis behavior

reneleonhardt:
## What does this PR do?

Bundles four related behavior changes that touch the same scan paths, and settles three contracts for how the scanner reports provenance:

- fix(mcp): Bound cancellable project scans — propagates request cancellation through scanner, Git, handoff, and MCP traversals, and bounds project and manifest discovery.
- feat(analysis): Add structured dependency contract — adds versioned, deterministic dependency JSON and MCP structured output for agentic coding in isolated worktrees and restricted sandboxes while preserving the existing text response; coverage remains explicit, including known partial Rust analysis.
- feat(scanner): Fail closed and report fallback provenance — prevents failed, timed-out, or unavailable ast-grep scans from appearing complete, and carries Cargo metadata fallback provenance through file graphs.
- feat(scanner): Resolve JS/TS workspace imports — resolves local JS/TS imports from package, pnpm, Bun, and Deno workspace manifests; ambiguous, external, escaping, and unsupported targets remain unresolved.

### 1. One provenance vocabulary

`analysis` owns the shared contract (`analysis.SourceStatus`, `analysis.Source`, `analysis.Coverage`, `analysis.NormalizeCoverage`); the scanner consumes it rather than defining a parallel copy.
The byte-identical `scanner.ScanSourceStatus`/`ScanSourceOutcome` duplicate is gone — what remains are aliases (`type ScanSourceStatus = analysis.SourceStatus`, `type ScanSourceOutcome = analysis.Source`), so there is exactly one enum and one struct in the binary.
Cargo-metadata and the new `rust-cargo` (mixed) source feed the same `analysis.Source` list.

### 2. One scanner entry point

`ScanForDeps(ctx, root, filters) (ScanOutcome, error)` collapses the WithFilters × Context × Outcome axes.
All retired compatibility twins were removed (`ScanForDepsContext`, `ScanForDepsWithFilters`, `ScanForDepsOutcome*`, `BuildFileGraphContext`, `BuildFileGraphWithFilters*`, `BuildFileGraphFrom*Context/FilteredAnalyses/OutcomeWithFilters`, `ScanDirectoryContext/Outcome`, `ScanConfiguredFilesContext`, `ScanFilesContext`, `ReadExternalDepsContext`, `GitDiffInfoContext`, `GitDiffFiles*`, `AnalyzeImpactContext`, `DependencyCoverageContext`, `HasConfiguredLanguageContext`).
`codemap` has no external importers, so the compatibility surface was dropped rather than preserved; all callers (CLI, MCP, handoff, blast-radius, watch, render, cmd) were updated to the single API, and ctx is threaded through every subprocess-bounded path.

### 3. Fail-closed without losing the degraded path

A timed-out or failed ast-grep scan now fails closed to an honest, usable answer instead of a hard error:

- `ScanOutcome{Analyses: nil, Sources: [{name: "ast-grep", status: timeout|failed, detail}]}` is returned with a nil error, so hooks (which run codemap on every edit) and MCP keep a usable empty answer on exactly the large repos where a 30s timeout fires — previously that was a hard error.
- A genuinely unavailable scanner (ast-grep not installed) still returns `ErrAstGrepNotFound` wrapped in `IncompleteScanError`, so setup problems stay visible.
- Deps coverage (`--deps --json`, MCP `get_dependencies`) is derived from the graph's source provenance via `CoverageFromSources`, so a degraded scan is observable in the output (`unavailable`/`partial`) instead of silently looking complete; Rust repos report `partial` through the `rust-cargo` mixed source, and cargo-metadata fallback provenance is surfaced the same way.

## CLI / MCP surface

No new CLI commands or arguments, and no new MCP tools — the MCP tool set is unchanged (15 tools; `get_dependencies` merely gains an `OutputSchema`).
The changes affect the output and behavior of existing commands:

- `codemap --json --deps .` output gains `schema_version` and `coverage` (deterministic, versioned shape; MCP `get_dependencies` returns the same as structured content).
- `codemap --deps <path>` resolves local JS/TS imports from package, pnpm, Bun, and Deno workspace manifests.
- Failed or timed-out scans degrade to an empty outcome with `Sources` provenance instead of a hard error.
- Scan, Git, handoff, and MCP traversals honor request cancellation.

### DIFF
--- a/analysis/contracts.go
+++ b/analysis/contracts.go
@@ -1,0 +1,77 @@
+package analysis
+
+import (
+	"slices"
+	"sort"
+)
+
+const SchemaVersion = "codemap.analysis/v1"
+
+type CoverageStatus string
+
+const (
+	CoverageComplete    CoverageStatus = "complete"
+	CoveragePartial     CoverageStatus = "partial"
+	CoverageUnavailable CoverageStatus = "unavailable"
+)
+
+type SourceStatus string
+
+const (
+	SourceAuthoritative SourceStatus = "authoritative"
+	SourceMixed         SourceStatus = "mixed"
+	SourceFallback      SourceStatus = "fallback"
+	SourceTimeout       SourceStatus = "timeout"
+	SourceUnavailable   SourceStatus = "unavailable"
+	SourceFailed        SourceStatus = "failed"
+)
+
+type Source struct {
+	Name   string       `json:"name"`
+	Status SourceStatus `json:"status"`
+	Detail string       `json:"detail,omitempty"`
+}
+
+type Issue struct {
+	Code       string   `json:"code"`
+	Severity   string   `json:"severity,omitempty"`
+	Path       string   `json:"path,omitempty"`
+	Line       int      `json:"line,omitempty"`
+	Message    string   `json:"message"`
+	Candidates []string `json:"candidates,omitempty"`
+}
+
+type Coverage struct {
+	Status  CoverageStatus `json:"status"`
+	Sources []Source       `json:"sources"`
+	Issues  []Issue        `json:"issues"`
+}
+
+func NormalizeCoverage(coverage Coverage) Coverage {
+	coverage.Sources = slices.Clone(coverage.Sources)
+	coverage.Issues = slices.Clone(coverage.Issues)
+	if coverage.Sources == nil {
+		coverage.Sources = []Source{}
+	}
+	if coverage.Issues == nil {
+		coverage.Issues = []Issue{}
+	}
+	sort.Slice(coverage.Sources, func(i, j int) bool {
+		left, right := coverage.Sources[i], coverage.Sources[j]
+		if left.Name != right.Name {
+			return left.Name < right.Name
+		}
+		if left.Status != right.Status {
+			return left.Status < right.Status
+		}
+		return left.Detail < right.Detail
+	})
+	for index := range coverage.Issues {
+		coverage.Issues[index].Candidates = slices.Clone(coverage.Issues[index].Candidates)
+		if coverage.Issues[index].Candidates == nil {
+			coverage.Issues[index].Candidates = []string{}
+		}
+		sort.Strings(coverage.Issues[index].Candidates)
+	}
+	return coverage
+}
--- a/analysis/contracts_test.go
+++ b/analysis/contracts_test.go
@@ -1,0 +1,45 @@
+package analysis
+
+import (
+	"encoding/json"
+	"reflect"
+	"testing"
+)
+
+func TestNormalizeCoverageProducesDeterministicNonNilCollections(t *testing.T) {
+	coverage := NormalizeCoverage(Coverage{
+		Status: CoveragePartial,
+		Sources: []Source{
+			{Name: "zeta", Status: SourceFallback},
+			{Name: "alpha", Status: SourceAuthoritative},
+		},
+		Issues: []Issue{{Code: "ambiguous", Message: "choose", Candidates: []string{"z", "a"}}},
+	})
+	if got, want := []string{coverage.Sources[0].Name, coverage.Sources[1].Name}, []string{"alpha", "zeta"}; !reflect.DeepEqual(got, want) {
+		t.Fatalf("source order = %v, want %v", got, want)
+	}
+	if got, want := coverage.Issues[0].Candidates, []string{"a", "z"}; !reflect.DeepEqual(got, want) {
+		t.Fatalf("candidates = %v, want %v", got, want)
+	}
+	encoded, err := json.Marshal(NormalizeCoverage(Coverage{}))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(encoded) == "" || string(encoded) == `{"status":"","sources":null,"issues":null}` {
+		t.Fatalf("coverage must encode non-null collections: %s", encoded)
+	}
+}
+
+func TestNormalizeCoveragePreservesIssueOrderAndSortsRepeatedFields(t *testing.T) {
+	input := Coverage{
+		Sources: []Source{{Name: "same", Status: SourceFallback}, {Name: "same", Status: SourceAuthoritative}},
+		Issues:  []Issue{{Code: "z", Candidates: []string{"b", "a"}}, {Code: "a"}},
+	}
+	got := NormalizeCoverage(input)
+	if got.Sources[0].Status != SourceAuthoritative {
+		t.Fatalf("source tie order = %#v", got.Sources)
+	}
+	if got.Issues[0].Code != "z" || !reflect.DeepEqual(got.Issues[0].Candidates, []string{"a", "b"}) || got.Issues[1].Code != "a" {
+		t.Fatalf("issue order = %#v", got.Issues)
+	}
+}
--- a/blast_radius.go
+++ b/blast_radius.go
@@ -17,6 +17,7 @@ import (
 	"strings"
 	"unicode/utf8"
 
+	"codemap/analysis"
 	"codemap/config"
 	"codemap/render"
 	"codemap/scanner"
@@ -68,6 +69,8 @@ type blastRadiusDeps struct {
 	ExternalDeps      map[string][]string    `json:"external_deps"`
 	DiffRef           string                 `json:"diff_ref,omitempty"`
 	ChangedFilesTotal int                    `json:"changed_files_total"`
+	CoverageStatus    string                 `json:"coverage_status,omitempty"`
+	CoverageNotes     []string               `json:"coverage_notes,omitempty"`
 }
 
 type blastRadiusImporters struct {
@@ -497,12 +500,16 @@ func buildBlastRadiusBundle(absRoot, ref string, limits blastRadiusLimits) (blas
 			return blastRadiusBundle{}, err
 		}
 		depsTotal = len(depsProject.Files)
-		depsCapped = capBlastRadiusDepsProject(depsProject, limits.MaxChangedFiles)
 
 		fg, err := scanner.BuildFileGraphFromOutcome(context.Background(), absRoot, scanOutcome, filters)
 		if err != nil {
 			return blastRadiusBundle{}, err
 		}
+
+		// Carry scan provenance into the deps project so a fail-closed scan
+		// never reads as an empty, complete dependency graph.
+		depsProject.Coverage = scanner.CoverageFromSources(fg.Coverage.Sources)
+		depsCapped = capBlastRadiusDepsProject(depsProject, limits.MaxChangedFiles)
 
 		// Analyze the FULL changed set; capping only limits what is displayed,
 		// it must not shrink blast-radius analysis or the summary stats.
@@ -549,6 +556,8 @@ func buildBlastRadiusBundle(absRoot, ref string, limits blastRadiusLimits) (blas
 			ExternalDeps:      depsCapped.ExternalDeps,
 			DiffRef:           depsCapped.DiffRef,
 			ChangedFilesTotal: depsTotal,
+			CoverageStatus:    string(depsCapped.Coverage.Status),
+			CoverageNotes:     coverageNotesFromSources(depsCapped.Coverage.Sources),
 		},
 		Importers:                    jsonReports,
 		Limits:                       limits,
@@ -601,6 +610,18 @@ func capBlastRadiusDepsProject(project scanner.DepsProject, max int) scanner.Dep
 	}
 	project.Files = append([]scanner.FileAnalysis(nil), project.Files[:max]...)
 	return project
+}
+
+// coverageNotesFromSources flattens source-level detail into coverage notes,
+// mirroring how the graph records blind-spot details for importers.
+func coverageNotesFromSources(sources []analysis.Source) []string {
+	var notes []string
+	for _, source := range sources {
+		if source.Detail != "" {
+			notes = append(notes, source.Detail)
+		}
+	}
+	return notes
 }
 
 func capBlastRadiusImportersReport(report scanner.ImportersReport, max int) blastRadiusImporters {
@@ -1277,7 +1298,7 @@ func renderDiffProject(project scanner.Project) string {
 
 func renderDepsProject(project scanner.DepsProject) string {
 	var buf bytes.Buffer
-	render.Depgraph(&buf, project)
+	render.Depgraph(context.Background(), &buf, project)
 	return stripANSI(buf.String())
 }
 
--- a/blast_radius.go
+++ b/blast_radius.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -427,7 +428,7 @@ func resolveBlastRadiusRoot(root string) (string, func(), error) {
 }
 
 func buildBlastRadiusBundle(absRoot, ref string, limits blastRadiusLimits) (blastRadiusBundle, error) {
-	diffInfo, err := scanner.GitDiffInfo(absRoot, ref)
+	diffInfo, err := scanner.GitDiffInfo(context.Background(), absRoot, ref)
 	if err != nil {
 		return blastRadiusBundle{}, &blastRadiusDiffError{ref: ref, err: err}
 	}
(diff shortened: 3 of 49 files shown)


JordanCoin:
@codex review

Focus areas for this one, since it consolidates four previously-separate PRs and touches every scan path:

1. **Cancellation correctness** — ctx is now threaded through scanner, git, handoff, and MCP. Check for paths that ignore ctx, leak subprocesses on cancel, or return partial results as if complete.
2. **The fail-closed contract** — a timed-out/failed ast-grep scan is supposed to degrade to `ScanOutcome{Analyses: nil, Sources: [...]}` with a **nil** error, while a genuinely unavailable scanner still errors. Verify no caller re-hardens that into a fatal path, and that degraded coverage is actually observable in output rather than silently looking complete.
3. **JS/TS workspace resolution** (`scanner/jsworkspace.go`, ~976 new lines) — exports/imports maps, pnpm/Bun/Deno workspaces, tsconfig rootDir/outDir remapping. Look for wrong-package resolution, escaping-path handling, symlink/case-sensitivity issues, and Windows path separator handling.
4. **Coverage provenance** — `analysis.Coverage` is now derived from graph source provenance via `CoverageFromSources`. Check the status derivation (complete/partial/unavailable) is correct for mixed-source repos.

Cross-platform behavior matters here: codemap ships on Windows, Linux, and macOS.

reneleonhardt:
All nine inline comments are addressed in commit "fix: Address maintainer review comments on #105".
Per-comment resolution with the regression test covering it:

- **Comment 01 (Copilot) + Comment 03 (P2)** — `render/depgraph.go`. `Depgraph` no longer re-runs `BuildFileGraph` (which re-triggered the ast-grep scan) on a project the caller already analyzed. It now takes a `context.Context` and builds the render graph from the caller's analyses via `BuildFileGraphFromAnalyses`, so MCP rendering observes request cancellation instead of launching a second unobservable scan with `context.Background()`. Covered by `TestDepgraph*` cases in `render/depgraph_test.go`.
- **Comment 02 (Copilot)** — `main.go` `runDepsMode`. The `--deps` graph build is now best-effort: a `BuildFileGraphFromOutcome` error no longer hard-exits; coverage falls back to the scan outcome's own sources when no graph is available.
- **Comment 04 (P2)** — `scanner/outcome.go`. `GraphCoverage.AddSource` no longer promotes every non-authoritative source to `partial`. Timeout/failed-only coverage stays `unavailable`; only a usable fallback or mixed source promotes to `partial`. Covered by `scanner/outcome_test.go`.
- **Comment 05 (P2)** — `scanner/jsworkspace.go` `parsePackageImports`. Conditional `#imports` targets (e.g. `{"#env": {"default": ..., "types": ...}}`) resolve through the same runtime-target selection used for `exports` instead of being invalidated as non-string mappings. Covered by the new `TestParsePackageImportsSkipsNonPackageKeys` (100% coverage on `parsePackageImports`).
- **Comment 06 (P2)** — `main.go` `runDepsFromStdin`. An empty stdin manifest now returns a valid empty `FileGraph` instead of a nil graph, so `graph.Coverage.Sources` no longer panics and the output stays the prior empty answer.
- **Comment 07 (P2)** — `blast_radius.go`. The blast-radius deps path keeps the scan outcome's sources and derives coverage for the deps output (`coverage_status`/`coverage_notes`), so a fail-closed scan is visible instead of looking like there were simply no changed dependencies.
- **Comment 08 (P2)** — `scanner/jsworkspace.go`. `tsconfig.json` `rootDir`/`outDir` inherited via `extends` is merged before `parseTSOutputDirs`, with a cyclic-extends guard so self-referential chains terminate. Covered by `TestMergedTSOutputDirs*` in `scanner/jsworkspace_test.go`.
- **Comment 09 (P2)** — `main.go` + `render/depgraph.go`. The non-JSON `--deps` path now renders `project.Coverage` via the new `renderCoverageLine`, so a timed-out scan shows `Coverage: unavailable/partial` (with source details) in text output instead of reading as a complete empty graph. Covered by the new `TestDepgraphRendersPartialCoverageWithoutDetail` (100% coverage on `renderCoverageLine`).

Full test suite green (`go test ./render ./scanner`, `-race` clean locally); `parsePackageImports` and `renderCoverageLine` each at 100% coverage.

reneleonhardt:
Adversarial review of the merged changes against a deep codemap turned up three edge cases the reviewers did not flag; all are fixed in commit "fix: Harden coverage ordering, stdin containment, JS specifiers") with focused unit tests:

- **Coverage ordering (extends Comment 04)** — `scanner/outcome.go` `GraphCoverage.AddSource` was order-dependent: recording a degraded source after a usable one (or a usable source after a degraded-only one) could leave the graph reading `unavailable` even though dependency references were extracted. `AddSource` is now order-independent: any usable source (authoritative/fallback/mixed) keeps coverage usable, and only all-degraded source sets read `unavailable`. Covered by `TestGraphCoverageAuthoritativeRepairsUnavailable` and `TestGraphCoverageDegradedAfterUsableKeepsPartial`.
- **stdin path containment (adjacent to Comment 06)** — `main.go` `runDepsFromStdin` wrote manifest files via `filepath.Join(tempDir, f.Path)` with no containment check, so a `--stdin` manifest declaring `"../x"` or an absolute path could write outside the private temp directory. `safeStdinManifestPath` now rejects absolute and `..`-traversing paths before any write. Covered by `TestSafeStdinManifestPath` and `TestRunDepsFromStdinRejectsEscapingPaths`.
- **Empty-scope JS specifiers (adjacent to Comment 05)** — `scanner/jsworkspace.go` `splitJSPackageSpecifier` accepted `"@/pkg"` as a scoped package even though the scope is empty, making a malformed specifier resolvable instead of failing closed. Covered by the `TestSplitJSPackageSpecifier` table.

Also mini-maximized unit coverage for the important changed paths (JS workspace `exports`/`imports`/target resolution, `resolveTarget` outDir→sourceRoot mapping, `newDepsProject` Rust detection, `ScanConfiguredFiles` `.codemap` exclusion, `summarizeEvents`/`summarizeHubs`, `gitCurrentBranchContext` error paths).

JordanCoin:
Re-reviewed at `861c630`. Verified your fixes rather than taking the commit messages at face value:

- **Depgraph second scan** — gone. `BuildFileGraphFromAnalyses` from caller analyses, ctx threaded. ✅
- **`--deps` hard exit** — now best-effort; coverage falls back to the outcome's sources. ✅
- **Stdin containment** — tested with both `../`-escape and absolute-path manifests. Both rejected, nothing written outside the temp dir. ✅
- Full suite green, `vet`/`staticcheck`/`gofmt` clean, builds on windows/amd64, linux/amd64, darwin/arm64.

**The stdin fix is more important than the commit message suggests.** I checked whether `main` has the same hole, and it does — a manifest with `../` escapes writes outside the temp dir on the currently shipped binary:

```
$ printf '{"files":[{"path":"../../../../../../../tmp/pwned.txt","content":"PWNED"}]}' | codemap --deps --stdin
$ cat /tmp/pwned.txt
PWNED
```

That's an arbitrary-file-write primitive in a mode agents drive with content they don't always control. Your branch rejects both forms. Worth calling out explicitly in the PR body so it doesn't land as a footnote — and it's a reason not to let this sit.

---

Two findings still open. My fault — I raised these in review notes but never actually posted them here, so you had no way to act on them. Both are in the coverage line, both cosmetic-but-user-facing:

**1. The Rust note prints twice.** Current output on a two-crate workspace:

```
Coverage: partial — Rust macro-generated, string-routed, and #[path] module edges may be
unresolved; Rust macro-generated, string-routed, and #[path] module edges may be unresolved
```

`ast-grep` and `rust-cargo` both carry an identical `Detail`, and `AddSource` appends each to `Notes` with no dedup. `NormalizeCoverage` sorts sources but doesn't dedupe notes.

**2. `ast-grep` is labeled `mixed` carrying a Rust-specific detail** (`astgrep.go`, the `DetectLanguage(...) == "rust"` loop). Its extraction isn't degraded — the macro/`#[path]`/string-routed caveat is a *Rust resolution* gap, which is `rust-cargo`'s to own. Two consequences:

- On a Go+Rust monorepo, the Go analysis gets reported as degraded too.
- It's what produces the duplicate in (1), since the same sentence ends up on two sources.

JSON today:

```json
{"name": "ast-grep",       "status": "mixed",         "detail": "Rust macro-generated, ..."}
{"name": "cargo-metadata", "status": "authoritative"}
{"name": "rust-cargo",     "status": "mixed",         "detail": "Rust macro-generated, ..."}
```

If `ast-grep` stays `authoritative` (it did its job) and only `rust-cargo` carries the Rust caveat, both issues resolve at once and the source list reads as what each tool actually contributed.

Neither blocks merge from my side — say the word if you'd rather I take them as a follow-up PR instead of spending your quota on it.

JordanCoin (approved):
Approving. Verified at `861c630` rather than trusting the commit messages:

**The three contracts hold under inspection**
- One vocabulary — `scanner.ScanSourceStatus` is a real alias to `analysis.SourceStatus`; the enum exists once.
- One entry point — grepped all 11 retired twins: zero definitions remain. `scanner/api_surface_test.go` parses the package AST and fails if any come back, so the contract is enforced rather than promised. Nice touch.
- Fail-closed with a degraded path — timeout/failure returns a usable `ScanOutcome` with `Sources` provenance and a nil error; genuinely-unavailable still errors so setup problems stay visible.

**Review fixes verified**
- Depgraph no longer re-scans; builds from caller analyses with ctx threaded.
- `--deps` graph build is best-effort; coverage falls back to the outcome's sources.
- Stdin containment tested with both `../`-escape and absolute-path manifests — both rejected, nothing written outside the temp dir.

**Verification**: full suite green, `vet`/`staticcheck`/`gofmt` clean, builds on windows/amd64, linux/amd64, darwin/arm64. Behavior spot-checked on real fixtures — npm workspace cross-package resolution (which `main` cannot do at all), Rust dev-dependency edges, and the coverage JSON contract.

**Scope check**: this is a strict superset of #93/#96/#97/#99 — all 30 of their files are present, none dropped. Of the 49 files, 19 are unique to this PR and account for only +549/-84: 12 test files plus 7 callers that had to migrate off the deleted twins. That migration is precisely why the four couldn't land independently.

The two coverage findings I raised are cosmetic and I'm not holding this on them — filing as issues instead. Thanks for the stdin catch; that one was live in the shipped binary.

reneleonhardt:
Both findings from your 2026-08-06 re-review are fixed on `fix/maintainer-contract-closure` (commit `19687f2` "fix(scanner): Keep ast-grep authoritative, dedupe coverage notes"), and the stdin fix is now called out in the PR body:

- **Finding 1 (duplicate Rust note)** — `scanner/outcome.go` `GraphCoverage.AddSource` now records each source detail once (dedup before append), and `render/depgraph.go` `renderCoverageLine` renders each detail once, so a shared caveat prints a single warning regardless of how many sources carry it. Every `Notes` consumer (MCP structured `coverage_notes`, `--importers`, intent, watch state) inherits the dedup. Covered by `TestGraphCoverageAddSourceDedupesSharedDetail` and `TestDepgraphRendersSharedCoverageDetailOnce`.
- **Finding 2 (ast-grep mislabeled `mixed`)** — `scanner/astgrep.go` no longer demotes ast-grep when the scan contains Rust files; ast-grep stays `authoritative` (it did its job). The Rust caveat is owned by `rust-cargo` alone (the file graph's `.rs` loop and the default deps coverage mirror the same shape). A Go+Rust monorepo's Go analysis is no longer reported degraded, and the JSON reads as what each tool contributed:

```json
{"name": "ast-grep",       "status": "authoritative"}
{"name": "cargo-metadata", "status": "authoritative"}
{"name": "rust-cargo",     "status": "mixed",         "detail": "Rust macro-generated, ..."}
```

  Covered by `TestAstGrepScanDirectoryRustStaysAuthoritative` and `TestNewDepsProjectRustCaveatOwnedByRustCargoSource`.
- **PR body** — the `--deps --stdin` arbitrary-file-write fix is now a dedicated section in the PR description ("Security: `--deps --stdin` manifest path containment"), not a footnote.

Adversarial review of the changed lines found no remaining paths that attach the Rust caveat to ast-grep (`astgrep.go`, `types.go`, `filegraph.go` cross-checked), no races (the dedup helpers are per-call state; `slices.Contains` over tiny note lists), and the only behavioral consequence of the labeling change is intentional: when a Rust repo's graph build fails and coverage falls back to the bare scan outcome, it now reports the authoritative scan rather than a graph-level caveat that cannot be derived without the graph. Targeted `go vet ./scanner ./render ./analysis .` clean; `go test ./scanner`, render `TestDepgraph*`, and the coverage/contract tests pass.

```diff
diff --git a/render/depgraph.go b/render/depgraph.go
index 8cd493a..8e87ba7 100644
--- a/render/depgraph.go
+++ b/render/depgraph.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 	"unicode"
@@ -381,7 +382,10 @@ func renderCoverageLine(w io.Writer, coverage analysis.Coverage) {
 	}
 	var details []string
 	for _, source := range coverage.Sources {
-		if source.Detail != "" {
+		// Sources are normalized (sorted, deduped by name/status/detail), but
+		// two distinct sources can still carry the same caveat; render each
+		// detail once so the warning is not doubled.
+		if source.Detail != "" && !slices.Contains(details, source.Detail) {
 			details = append(details, source.Detail)
 		}
 	}
diff --git a/scanner/astgrep.go b/scanner/astgrep.go
index d7c9613..df522db 100644
--- a/scanner/astgrep.go
+++ b/scanner/astgrep.go
@@ -223,17 +223,15 @@ func (s *AstGrepScanner) ScanDirectory(parent context.Context, root string) (Sca
 		}
 		return ScanOutcome{}, err
 	}
-	source := analysis.Source{Name: "ast-grep", Status: analysis.SourceAuthoritative}
-	for _, fileAnalysis := range analyses {
-		if DetectLanguage(fileAnalysis.Path) == "rust" {
-			source.Status = analysis.SourceMixed
-			source.Detail = rustCoverageNote
-			break
-		}
-	}
+	// ast-grep extraction is authoritative regardless of language: it scanned
+	// the files and reported every reference it found. The Rust caveat
+	// (macro-generated, string-routed, and #[path] edges) is a *resolution*
+	// gap owned by rust-cargo, which the file graph records separately; labeling
+	// ast-grep mixed here would degrade a Go+Rust monorepo's Go analysis and
+	// duplicate the same caveat on two sources.
 	return ScanOutcome{
 		Analyses: analyses,
-		Sources:  []analysis.Source{source},
+		Sources:  []analysis.Source{{Name: "ast-grep", Status: analysis.SourceAuthoritative}},
 	}, nil
 }
 
diff --git a/scanner/outcome.go b/scanner/outcome.go
index f8295e2..d128d58 100644
--- a/scanner/outcome.go
+++ b/scanner/outcome.go
@@ -2,6 +2,7 @@ package scanner
 
 import (
 	"fmt"
+	"slices"
 
 	"codemap/analysis"
 )
@@ -62,7 +63,10 @@ func (c *GraphCoverage) AddSource(outcome ScanSourceOutcome) {
 			c.Status = analysis.CoverageUnavailable
 		}
 	}
-	if outcome.Detail != "" {
+	// Record the detail once: distinct sources may legitimately carry the same
+	// caveat (e.g. a shared Rust resolution note), and consumers surface every
+	// note, so a duplicate would read as a doubled warning in the output.
+	if outcome.Detail != "" && !slices.Contains(c.Notes, outcome.Detail) {
 		c.Notes = append(c.Notes, outcome.Detail)
 	}
 }
diff --git a/scanner/types.go b/scanner/types.go
index 859b82b..46d44ef 100644
--- a/scanner/types.go
+++ b/scanner/types.go
@@ -64,24 +64,34 @@ type DepsProject struct {
 // newDepsProject builds a DepsProject with default coverage derived from the
 // analyses and inventory (Rust repos are partial). Production callers pass
 // explicit graph-derived coverage via NewDepsProjectWithCoverage; this default
-// constructor is kept for the determinism and Rust-coverage tests.
+// constructor is kept for the determinism and Rust-coverage tests. It mirrors
+// the graph-derived production shape: ast-grep stays authoritative (it did its
+// job) and the Rust caveat is carried by a rust-cargo source, never attached
+// to ast-grep, so the source list reads as what each tool contributed.
 func newDepsProject(root string, files []FileAnalysis, externalDeps map[string][]string, diffRef string, inventory ...[]FileInfo) DepsProject {
 	coverage := analysis.Coverage{
 		Status:  analysis.CoverageComplete,
 		Sources: []analysis.Source{{Name: "ast-grep", Status: analysis.SourceAuthoritative}},
 	}
+	addRustCaveat := func() {
+		if coverage.Status == analysis.CoveragePartial {
+			return
+		}
+		coverage.Status = analysis.CoveragePartial
+		coverage.Sources = append(coverage.Sources, analysis.Source{
+			Name: "rust-cargo", Status: analysis.SourceMixed, Detail: rustCoverageNote,
+		})
+	}
 	for _, file := range files {
 		if file.Language == "rust" || DetectLanguage(file.Path) == "rust" {
-			coverage.Status = analysis.CoveragePartial
-			coverage.Sources[0].Detail = rustCoverageNote
+			addRustCaveat()
 			break
 		}
 	}
 	if coverage.Status == analysis.CoverageComplete && len(inventory) > 0 {
 		for _, file := range inventory[0] {
 			if DetectLanguage(file.Path) == "rust" {
-				coverage.Status = analysis.CoveragePartial
-				coverage.Sources[0].Detail = rustCoverageNote
+				addRustCaveat()
 				break
 			}
 		}
```

Issue #107, opened 2026-08-06
coverage: Rust caveat prints twice in the CLI coverage line


JordanCoin:
Split out of #105 review — cosmetic, not a correctness bug, deliberately not held against that PR.

## What happens

The same sentence is printed twice, joined by `; `:

```
$ codemap --importers helper/src/lib.rs <cargo-workspace>
📍 File: helper/src/lib.rs
   Imported by 1 file(s)
   • app/src/lib.rs
Coverage: partial — Rust macro-generated, string-routed, and #[path] module edges may be
unresolved; Rust macro-generated, string-routed, and #[path] module edges may be unresolved
```

## Why

Both the `ast-grep` and `rust-cargo` sources carry an identical `Detail`, and `GraphCoverage.AddSource` appends each one to `Notes` unconditionally:

```go
if outcome.Detail != "" {
    c.Notes = append(c.Notes, outcome.Detail)
}
```

`analysis.NormalizeCoverage` sorts `Sources` and clones `Issues[].Candidates`, but never dedupes `Notes`.

## Two ways to fix

1. **Fix the cause** — resolve #106 so only `rust-cargo` carries the Rust caveat. The duplicate disappears with no dedup logic needed. Preferred.
2. **Fix the symptom** — dedupe `Notes` in `AddSource` or `NormalizeCoverage`. Worth doing anyway as a belt-and-braces measure, since any two sources sharing a detail string will otherwise repeat it.

Doing (1) alone resolves the reported symptom; doing both makes it structurally impossible to regress.

## Reproduce

Cargo workspace with two crates where `app` uses `helper` from a `#[cfg(test)]` block, then run `codemap --importers helper/src/lib.rs <workspace>` and read the trailing `Coverage:` line.

JordanCoin:
Cross-ref: filed #111 for the underlying design issue — coverage is reported at scan granularity but consumed at query granularity, which is why this keeps surfacing. This issue is still worth fixing on its own; #111 is the reason it recurs.

Also reclassifying: I filed this as cosmetic assuming a human reader. If the consumer is an agent parsing `coverage.sources[]` / `coverage.notes[]`, a duplicated note is a malformed payload and a misattributed source is wrong data — not a presentation nit.

JordanCoin:
Fixed by #108 (`d1e8a13`), verified on merged main with a two-crate Cargo workspace:

- `ast-grep` now reports `authoritative`; only `rust-cargo` carries the Rust caveat.
[...]

Issue #106, opened 2026-08-06
coverage: ast-grep reports `mixed` carrying a Rust-specific caveat


JordanCoin:
Split out of #105 review — cosmetic, not a correctness bug, deliberately not held against that PR.

## What happens

`scanner/astgrep.go` marks the **ast-grep** source `mixed` with the Rust caveat whenever any analyzed file is Rust:

```go
source := analysis.Source{Name: "ast-grep", Status: analysis.SourceAuthoritative}
for _, fileAnalysis := range analyses {
    if DetectLanguage(fileAnalysis.Path) == "rust" {
        source.Status = analysis.SourceMixed
        source.Detail = rustCoverageNote
        break
    }
}
```

On a two-crate Cargo workspace that produces:

```json
{"name": "ast-grep",       "status": "mixed",         "detail": "Rust macro-generated, string-routed, and #[path] module edges may be unresolved"}
{"name": "cargo-metadata", "status": "authoritative"}
{"name": "rust-cargo",     "status": "mixed",         "detail": "Rust macro-generated, string-routed, and #[path] module edges may be unresolved"}
```

## Why it's wrong

ast-grep's extraction isn't degraded — it found what it found. The caveat (macro-generated, `#[path]`, string-routed edges) describes a **Rust resolution** gap, which is `rust-cargo`'s concern. Conflating "which tool ran" with "which language has resolution gaps" has two consequences:

1. On a **Go + Rust monorepo**, the Go analysis is reported as degraded too, even though nothing about the Go extraction is partial.
2. It's the direct cause of #107 — the same sentence lands on two sources and prints twice.

## Suggested fix

Leave `ast-grep` as `authoritative` and let only `rust-cargo` carry the Rust caveat. The source list then reads as what each tool actually contributed, and the duplicate note disappears at the same time.

A third site does the same thing: `scanner/types.go` in `newDepsProject` sets `coverage.Sources[0].Detail = rustCoverageNote` on the ast-grep source (kept "for the determinism and Rust-coverage tests").

## Reproduce

Cargo workspace with two crates where `app` uses `helper` from a `#[cfg(test)]` block, then:

```
codemap --json --deps <workspace> | jq .coverage
```

JordanCoin:
Cross-ref: filed #111 for the underlying design issue — coverage is reported at scan granularity but consumed at query granularity, which is why this keeps surfacing. This issue is still worth fixing on its own; #111 is the reason it recurs.

[...]

PR #108, opened 2026-08-07
fix(scanner): Keep ast-grep authoritative, dedupe coverage notes

reneleonhardt:
Follow-up to #105, closing the two open findings from the maintainer re-review on the coverage line.

## What this PR does

Both findings were in the coverage line — cosmetic but user-facing:

**1. The Rust note printed twice.** `ast-grep` and `rust-cargo` both carried an identical `Detail`, and `GraphCoverage.AddSource` appended each source's detail to `Notes` with no dedup. A two-crate workspace rendered:

```
Coverage: partial — Rust macro-generated, string-routed, and #[path] module edges may be
unresolved; Rust macro-generated, string-routed, and #[path] module edges may be unresolved
```

**2. `ast-grep` was labeled `mixed` carrying a Rust-specific detail.** The `DetectLanguage(...) == "rust"` loop in `scanner/astgrep.go` demoted ast-grep's source status even though its extraction is not degraded — the macro/`#[path]`/string-routed caveat is a *Rust resolution* gap, which `rust-cargo` owns. Consequences: on a Go+Rust monorepo the Go analysis was reported degraded too, and the same sentence landed on two sources, producing the duplicate in (1).

## Changes

- **`scanner/astgrep.go`** — ast-grep stays `authoritative` on a successful scan regardless of language; the Rust caveat is no longer attached to its source. The file graph already records `rust-cargo` (mixed, with the caveat) for Rust repos, so coverage remains `partial` where it should.
- **`scanner/outcome.go`** — `GraphCoverage.AddSource` records each source detail once (dedup before append), so every `Notes` consumer (MCP structured `coverage_notes`, `--importers`, intent, watch state) inherits the dedup.
- **`render/depgraph.go`** — `renderCoverageLine` renders each detail once, so a shared caveat prints a single warning in the text output even if two sources carry it.
- **`scanner/types.go`** — the default `newDepsProject` coverage mirrors the production shape (ast-grep `authoritative` + `rust-cargo` `mixed` carrying the note) instead of attaching the note to ast-grep.

## Before / after

JSON sources before:

```json
{"name": "ast-grep",       "status": "mixed",         "detail": "Rust macro-generated, ..."}
{"name": "cargo-metadata", "status": "authoritative"}
{"name": "rust-cargo",     "status": "mixed",         "detail": "Rust macro-generated, ..."}
```

JSON sources after — the list reads as what each tool actually contributed:

```json
{"name": "ast-grep",       "status": "authoritative"}
{"name": "cargo-metadata", "status": "authoritative"}
{"name": "rust-cargo",     "status": "mixed",         "detail": "Rust macro-generated, ..."}
```

## Regression tests

- `TestAstGrepScanDirectoryRustStaysAuthoritative` — Rust scans keep ast-grep `authoritative` with no Rust detail.
- `TestGraphCoverageAddSourceDedupesSharedDetail` — shared caveat recorded once across sources.
- `TestDepgraphRendersSharedCoverageDetailOnce` — a shared detail renders once in text output.
- `TestNewDepsProjectRustCaveatOwnedByRustCargoSource` — default coverage shape matches production.

## Verification

- `go build ./...`, `go vet ./scanner ./render ./analysis .`, `gofmt` clean.
- Targeted `scanner` coverage/contract tests and render `TestDepgraph*` pass.

Co-authored-by: GPT-5.6 Sol <codex@openai.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ __pycache__/
 
 # Runtime state
 .codemap/
+.cache/
 /codemap-dev
 /dev_codemap
 /bundled-tools/
--- a/render/depgraph.go
+++ b/render/depgraph.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 	"unicode"
@@ -381,7 +382,10 @@ func renderCoverageLine(w io.Writer, coverage analysis.Coverage) {
 	}
 	var details []string
 	for _, source := range coverage.Sources {
-		if source.Detail != "" {
+		// Sources are normalized (sorted, deduped by name/status/detail), but
+		// two distinct sources can still carry the same caveat; render each
+		// detail once so the warning is not doubled.
+		if source.Detail != "" && !slices.Contains(details, source.Detail) {
 			details = append(details, source.Detail)
 		}
 	}
--- a/render/depgraph.go
+++ b/render/depgraph.go
@@ -382,9 +382,9 @@ func renderCoverageLine(w io.Writer, coverage analysis.Coverage) {
 	}
 	var details []string
 	for _, source := range coverage.Sources {
-		// Sources are normalized (sorted, deduped by name/status/detail), but
-		// two distinct sources can still carry the same caveat; render each
-		// detail once so the warning is not doubled.
+		// NormalizeCoverage sorts Sources by name/status/detail but does not
+		// deduplicate them; two distinct sources can still carry the same
+		// caveat, so render each detail once to keep the warning single.
 		if source.Detail != "" && !slices.Contains(details, source.Detail) {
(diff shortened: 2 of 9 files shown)


JordanCoin (approved):
Verified on a real two-crate Cargo workspace rather than from the description:

**#106** — `ast-grep` is now `authoritative`; only `rust-cargo` carries the Rust caveat. Coverage still correctly reports `partial`, so the honesty is preserved while the attribution is fixed.

**#107** — the coverage line prints the caveat once.

No regression on a Go-only repo: still `complete` with a single authoritative `ast-grep` source.

Deduping at `AddSource` rather than at the renderer is the right call — every consumer (MCP `coverage_notes`, `--importers`, intent, watch state) inherits it, instead of each render path needing its own fix. Full suite green, vet clean. Thanks for turning these around before I'd even finished filing them.

PR #118, opened 2026-08-08
fix(scanner): Resolve Rust path-attributed modules

reneleonhardt:
## What does this PR do?

Resolve direct Rust `#[path = "..."]` modules and logical module ownership
across explicit target roots and `mod.rs` layouts. Ambiguous, malformed,
conditional, inline, macro-generated, and out-of-repository routes remain
conservative and are reported as partial coverage.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

## Additional notes

Focused path, ownership, ambiguity, malformed-input, and cancellation tests,
`go vet ./...`, the full race/coverage suite on macOS Go 1.26.5, and the full
suite on Linux Go 1.24.13 are GREEN.

### DIFF
--- a/scanner/astgrep.go
+++ b/scanner/astgrep.go
@@ -324,14 +324,23 @@ func (s *AstGrepScanner) scanDirectory(parent context.Context, root string) ([]F
 			}
 		}
 
-		if m.RuleID == "rust-mod-imports" || m.RuleID == "rust-path-imports" || m.RuleID == "rust-use-imports" {
+		if m.RuleID == "rust-mod-imports" || m.RuleID == "rust-path-module-imports" || m.RuleID == "rust-path-imports" || m.RuleID == "rust-use-imports" {
 			var path string
+			var explicitTarget string
 			kind := "rust-path"
 			switch m.RuleID {
 			case "rust-mod-imports":
 				kind = "rust-module"
 				if pathVar, ok := m.MetaVariables.Single["PATH"]; ok {
 					path = pathVar.Text
+				}
+			case "rust-path-module-imports":
+				kind = "rust-path-module"
+				if pathVar, ok := m.MetaVariables.Single["PATH"]; ok {
+					path = pathVar.Text
+				}
+				if targetVar, ok := m.MetaVariables.Single["TARGET"]; ok {
+					explicitTarget = targetVar.Text
 				}
 			case "rust-path-imports":
 				path = m.Text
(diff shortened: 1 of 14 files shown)


JordanCoin:
Heads-up before these land: **#117 and #118 don't compile together**, in either merge order. Both are green in CI because each is built against `main`, not against the other.

Git auto-merges them with no textual conflict — the collision is semantic:

`main` today:
```go
func discoverCargoManifests(root string, files []FileInfo) []string
```

**#118** changes the signature:
```go
func discoverCargoManifests(ctx context.Context, root string, files []FileInfo) ([]string, error)
```

**#117** adds a new caller (`scanner/cargofallback.go:72`) using the old one:
```go
manifests := discoverCargoManifests(root, files)
```

Result whichever way round they go:

```
scanner/cargofallback.go:72:15: assignment mismatch: 1 variable but discoverCargoManifests returns 2 values
scanner/cargofallback.go:72:44: not enough arguments in call to discoverCargoManifests
	have (string, []FileInfo)
	want (context.Context, string, []FileInfo)
```

Verified by merging both into a scratch worktree in each order and building.

The fix is one line in `buildCargoFallbackOutcome`, which already has a `ctx` in scope:

```go
manifests, err := discoverCargoManifests(ctx, root, files)
if err != nil {
    return ScanOutcome{}, err
}
```

Simplest is to fix it in whichever of the two you'd like to merge second. No need to change anything else — #115, #116 and #119 are unaffected, and all five merge cleanly apart from this.

JordanCoin (changes requested):
This implementation is sound in isolation, but #117 has now landed and the combined tree does not compile.  still calls , while this PR changes it to . Applying the two changes in either order reproduces the assignment mismatch/not-enough-arguments compiler error. Please update that caller, propagate the error consistently, and rerun the scanner suite on current main; I will then re-review and merge.

JordanCoin:
Correction with the exact symbols omitted by shell formatting in my review: merged #117 adds `scanner/cargofallback.go`, which calls `discoverCargoManifests(root, files)`. This PR changes the function to `discoverCargoManifests(ctx, root, files) ([]string, error)`. The combined compiler error is an assignment mismatch plus not enough arguments.

JordanCoin (approved):
Re-reviewed at 2f56bce after #117 integration. The Cargo fallback caller now uses context-aware manifest discovery, propagates cancellation, and has a focused regression test. The exact combined tree passes the focused scanner suite, full local test suite, and all 12 CI jobs. Approved.

PR #121, opened 2026-08-12
fix(mcp): Align context and status surfaces

reneleonhardt:
## What does this PR do?

Keeps MCP results honest and deterministic: ordinary text is bounded to the output budget, legacy JSON responses stay complete, project paths resolve to native/request-local roots (including `~` home expansion), and the status surface is stable and exact. The tool inventory is served by the protocol's `tools/list` (the SDK's live registry) instead of a hand-maintained embedded list that could drift from the registered set.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

## Additional notes

- Bounds ordinary text and status output to `limits.MaxContextOutputBytes`; `completeJSONResult` preserves complete, valid JSON.
- Deterministic status/watcher/activity ordering; exact-surface contract asserted against the protocol `tools/list`.
- Native, request-local project roots; `~/` paths expand via `$HOME`.
- ~500 lines of exhaustive surface tests (call-site classification, bounding, determinism, inventory, request-local roots).

### DIFF
--- a/mcp/analysis_output.go
+++ b/mcp/analysis_output.go
@@ -1,6 +1,8 @@
 package codemapmcp
 
 import (
+	"sort"
+
 	"codemap/handoff"
 	"codemap/scanner"
 )
@@ -78,6 +80,9 @@ func newImportersOutput(root, file string, graph *scanner.FileGraph) ImportersOu
 			hubImports = append(hubImports, imported)
 		}
 	}
+	sort.Strings(importers)
+	sort.Strings(imports)
+	sort.Strings(hubImports)
 	boundedImporters, omittedCount := boundedSlice(importers)
 	boundedImports, omittedCount := addBounded(imports, omittedCount)
 	boundedHubs, omittedCount := addBounded(hubImports, omittedCount)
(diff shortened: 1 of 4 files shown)


JordanCoin (approved):
Verified independently. The `status` tool was advertising 10 tools while 16 are registered — deleting that hand-maintained list is a real truthfulness fix, and a list like that could only ever drift. Confirmed the new tests aren't vacuous: reverted the source to main and 8 of 10 fail with the exact symptoms claimed.

Noted for follow-up, not blocking: `mcp/main.go:334` uses `os.Getenv("HOME")` for `~` expansion, which is unset on Windows (`os.UserHomeDir()` reads `USERPROFILE`), so `~/proj` resolves against the server's cwd. It's propagation rather than regression — 8 such sites on main, 9 here — and `skills/loader.go:40` already has the right idiom. Worth knowing that CI's test matrix is ubuntu+macos only, so this class is invisible to CI.

PR #126, opened 2026-08-12
feat(scanner): Resolve Askama template dependencies

reneleonhardt:
## What does this PR do?

Record file dependencies introduced by literal Askama `#[template(path = "...")]` attributes. Paths reuse the existing Rust literal decoder and Cargo package ownership, resolve under the package's default `templates/` directory, and are accepted only for one exact configured, indexed in-repository target. Dynamic, malformed, missing, external, excluded, escaping, ambiguous, and unowned routes remain unresolved. Packages with `askama.toml` and attributes with `config = ...` also fail closed; custom template-directory configuration is intentionally partial coverage.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

- [x] Documentation is unchanged because the dependency output is self-describing.

## Additional notes

Focused extraction and conservative-resolution tests, the scanner suite, and staticcheck are GREEN.

Real-world Worktrunk smoke tests recover all six template dependencies from `src/shell/mod.rs` to the Bash, Fish, Fish wrapper, Nushell, PowerShell, and Zsh files under `templates/`.

This branch targets and contains no generic proc-macro, Cargo-topology, configuration, or non-Rust changes.

### DIFF
--- a/scanner/astgrep.go
+++ b/scanner/astgrep.go
@@ -324,7 +324,7 @@ func (s *AstGrepScanner) scanDirectory(parent context.Context, root string) ([]F
 			}
 		}
 
-		if m.RuleID == "rust-mod-imports" || m.RuleID == "rust-path-module-imports" || m.RuleID == "rust-path-imports" || m.RuleID == "rust-use-imports" {
+		if m.RuleID == "rust-mod-imports" || m.RuleID == "rust-path-module-imports" || m.RuleID == "rust-path-imports" || m.RuleID == "rust-use-imports" || m.RuleID == "rust-askama-template-imports" {
 			var path string
 			var explicitTarget string
 			kind := "rust-path"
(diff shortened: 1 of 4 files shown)


JordanCoin (approved):
Verified on real Cargo crates. The property that makes this safe: resolution is existence-gated — an edge is only emitted when a file sits at exactly `<pkg>/templates/<literal>` and is uniquely indexed, with `files[0] != target` guarding against the index's extension-stripped keys. Tested the case that matters most: a crate with `askama.toml` pointing at `views/` plus a decoy at `templates/hello.html` produces **no edge at all** rather than guessing the decoy. `#[template(source = "...")]` with a same-named decoy file also correctly produces nothing.

On scope — it's a fair question whether codemap should carry per-crate support, and I think this earns it: `#[template(path=...)]` against `<crate>/templates/` is shared by askama, rinja, sailfish and yarte, so it's closer to a Rust convention than one library. And a special case that can only fail closed is much cheaper to own than one that guesses.

Two follow-ups, neither blocking: `path` must be the first attribute argument, so `#[template(escape = "none", path = "x.html")]` is silently missed (one more pattern fixes it); and on Windows `filepath.IsAbs("/foo.html")` is false, so a leading-slash literal resolves there while it's correctly rejected on Unix — a cross-platform divergence worth closing with a `strings.HasPrefix(value, "/")` check.

If a second template crate ever arrives, please table-drive it (attribute name, default dir, config filename) rather than adding a second resolver — the hard-coded `askama.toml` check is already the seam.

PR #128, opened 2026-08-12
fix(setup): Keep Codemap state locally ignored

reneleonhardt:
## What does this PR do?

Writes `.codemap/` to the clone-local `.git/info/exclude` (shared by linked worktrees, submodule-local) instead of a tracked `.gitignore`, so `setup` and `config init` never touch tracked files. `codemap doctor` verifies the effective ignore state: OK whenever git ignores `.codemap/` by any mechanism, MISS only when it is genuinely unignored, with a minimal repair hint.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

## Additional notes

- Appends to `info/exclude` only when the entry is missing: idempotent, no duplicate on re-run, never creates a `.gitignore`.
- Atomic write: a symlinked exclude is replaced, not written through; CRLF files and files without a trailing newline are preserved byte-for-byte.
- Repair hint runs `codemap setup --no-hooks --no-mcp --no-config` so fixing the ignore rule does not rewrite hooks or MCP config.
- Silent no-op outside a Git working tree.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -80,7 +80,7 @@ func configInit(root string) {
 	}
 
 	fmt.Printf("Created %s\n", result.Path)
-	reportEnsureIgnored(os.Stdout, root, ignoreTracked)
+	reportEnsureIgnored(os.Stdout, root)
 	fmt.Println()
 	if len(result.TopExts) == 0 {
 		fmt.Println("No code extensions detected — wrote empty config.")
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -112,6 +112,9 @@ func RunDoctor(args []string, defaultRoot string) int {
 	}
 
 	checkFile("project config", config.ConfigPath(root), validateJSONFile)
+	if isGitWorkTree(root) {
+		doctorCheckLocalCodemapIgnore(root, &failures)
+	}
 	claudeSettings, claudeSettingsErr := claudeSettingsPath(root, *global)
 	claudeMCP, claudeMCPErr := claudeMCPPath(root, *global)
 	codexHooks, codexHooksErr := codexHooksPath(root, *global)
@@ -172,6 +175,31 @@ func RunDoctor(args []string, defaultRoot string) int {
 	}
 	fmt.Println("\nCodemap integration prerequisites are valid.")
 	return 0
+}
+
+// doctorCheckLocalCodemapIgnore verifies .codemap/ is ignored in this working
+// tree. Git's check-ignore reflects the rules that actually apply, so it is
+// the authority: report OK when the directory is ignored and MISS only when
+// it genuinely is not, naming the local exclude and a minimal repair hint.
+func doctorCheckLocalCodemapIgnore(root string, failures *int) {
+	ignored, mechanism, err := gitCheckIgnoreVerbose(root, ignoreEntry)
+	if err != nil {
+		fmt.Printf("MISS local Codemap ignore: %v\n", err)
+		*failures++
+		return
+	}
+	if ignored {
+		fmt.Printf("OK   local Codemap ignore: effective via %s\n", mechanism)
+		return
+	}
+	path, _, err := localCodemapIgnore(root)
+	if err != nil {
+		fmt.Printf("MISS local Codemap ignore: %v\n", err)
+		*failures++
+		return
+	}
+	fmt.Printf("MISS local Codemap ignore: %s does not ignore %s; repair with `codemap setup --no-hooks --no-mcp --no-config %s`\n", path, ignoreEntry, root)
+	*failures++
 }
 
 func reportCodexRuntimeVersions(cliPath string) bool {
--- a/cmd/gitignore.go
+++ b/cmd/gitignore.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"bufio"
 	"bytes"
 	"fmt"
 	"io"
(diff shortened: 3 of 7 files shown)


JordanCoin:
Reviewed. The write path is genuinely careful and I like the core decision: it writes to `.git/info/exclude`, never to a tracked `.gitignore`. Setup is an install-time action, and appending to a committed `.gitignore` would create an unrequested diff for the whole team. `info/exclude` is also the lowest-precedence ignore source, so a team that deliberately tracks `.codemap/` still wins.

Verified across 15 disposable repos with byte-level before/after: fresh repo with no `.gitignore` (none created), existing `.gitignore` with unrelated entries (byte-identical after), already-ignored `.codemap/` (not duplicated), two consecutive runs (idempotent), pre-populated `exclude` (pure append), **no trailing newline** (no concatenation onto the last line), **CRLF** file (no corruption), linked worktree (lands in the main repo's exclude via `--git-common-dir`), submodule (lands in `.git/modules/sub/`), read-only exclude (warns, exits 0), and `GIT_DIR` pointing elsewhere (writes nothing, fails safe). That's thorough.

One thing I'd fix before merging.

## `cmd/doctor.go:115-126` — doctor fails repos where `.codemap/` is provably ignored

Reproduced on **codemap's own repo**:

```
$ git check-ignore -v .codemap/
.gitignore:12:.codemap/	.codemap/

$ codemap doctor --agent claude .
MISS local Codemap ignore: .../.git/info/exclude does not contain .codemap/;
     repair with `codemap setup ...`
1 check(s) need attention.
```

The runtime state *is* ignored — doctor just isn't looking at the mechanism that's doing it. Two consequences: every existing user configured by a prior version gets a brand-new failure and nonzero exit until they re-run setup, and this repo fails its own doctor.

It also contradicts the comment 200 lines further down the same file:

> *"Reporting it MISS because the project file does not repeat it describes the file layout rather than the effective configuration."*

That's the exact bug #100 fixed for hooks and MCP, reappearing as a new check. Consulting `gitCheckIgnore` first and reporting `OK` with the effective mechanism — reserving the failure for genuinely unignored repos — keeps the check useful without the false alarm. Or keep the note and don't `failures++`.

## Smaller

- `cmd/gitignore.go:180-188` — `hasIgnoreLine` uses `bufio.Scanner` and ignores `scanner.Err()`. On an `exclude` whose first line exceeds 64KB, `Scan()` returns false with `ErrTooLong` and `.codemap/` reads as absent even when present. Measured: the entry went 1 → 2 → 3 across two setup runs, and doctor then reports MISS permanently. Pre-existing function, but this PR puts it on setup's unconditional path *and* behind a new doctor failure. `bytes.Split` or `scanner.Buffer(...)` plus an `Err()` check.
- `cmd/gitignore.go:154-176` — `appendIgnoreEntry` uses `os.WriteFile`, which follows symlinks and is non-atomic. With `.git/info/exclude` symlinked outside the repo, setup **wrote through it**. The repo already has `writeFileAtomic` (`cmd/setup.go:38`) whose comment is about exactly this; `os.Rename` replaces the symlink instead of writing through it and removes the truncate-then-write window.
- `ignoreTracked` (`cmd/gitignore.go:19-21`) is now dead in production — both call sites moved to `ignoreLocal`, and only tests reference it, which is why staticcheck stays quiet.
- The notice prints a repo-escaping relative path in worktrees (`Added .codemap/ to ../k/.git/info/exclude`); prefer absolute when `filepath.Rel` starts with `..`.
- The repair hint suggests full `codemap setup`, which rewrites hooks and MCP to fix one ignore line; `--no-hooks --no-mcp --no-config` is the minimal repair.

Test gaps worth locking in, since all three pass today: CRLF, missing trailing newline in *local* mode (only tracked mode is covered), and a pre-populated `exclude`.

reneleonhardt:
Thanks for the review — all six points are addressed, with a couple of small extras.

**Confirmed, as reviewed:**
- Doctor no longer fails repos where `.codemap/` is provably ignored. It now consults `git check-ignore` first and reports `OK   local Codemap ignore: effective via <source>:<line>:<pattern>` (your exact repro now shows `effective via .gitignore:12:.codemap/`); MISS is reserved for genuinely unignored repos.
- `hasIgnoreLine` dropped `bufio.Scanner` for `bytes.Split` — a >64 KiB first line can no longer hide a present entry (regression test locked in).
- `appendIgnoreEntry` now writes via `writeFileAtomic` — symlinked `exclude` is replaced, not written through, and there's no truncate-then-write window.
- Dead `ignoreTracked` removed, along with the tracked-`.gitignore` write path it guarded.
- The notice keeps the path absolute when `filepath.Rel` would escape the root (linked worktrees).
- Repair hint is now the minimal `codemap setup --no-hooks --no-mcp --no-config`.

**Also changed:**
- Squashed to a single commit on top of current `main`, with a tightened commit message and PR body to match the final state.
- Tests added for the gaps you flagged: CRLF, missing trailing newline in *local* mode, and pre-populated `exclude`, plus the 70 KiB-line regression and a linked-worktree absolute-path notice.
- One behavioral consequence of the atomic write to flag: a read-only `exclude` in a writable `.git/info` is now silently replaced (rename) rather than warning. If you'd rather keep the warning, say so and I'll add an explicit writability probe.

JordanCoin (approved):
All three findings verified fixed.

Doctor now reports the effective state — `OK local Codemap ignore: effective via .gitignore:12:.codemap/` on this repo — and still MISSes a genuinely unignored repo, so the check wasn't gutted to make the failure go away. It also handles a non-git directory (gates off silently rather than reporting something misleading) and a repo with no commits.

`hasIgnoreLine` moved to `bytes.Split`, so a 70KB line in `.git/info/exclude` no longer makes an existing entry read as absent — idempotency holds at exactly one entry across repeated runs. And `appendIgnoreEntry` now goes through `writeFileAtomic`, so a symlinked `exclude` gets replaced rather than written through, and there's no truncate-then-write window.

The repair hint is also now the minimal `--no-hooks --no-mcp --no-config` form rather than a full setup. Thanks for taking all of it.

Issue #130, opened 2026-08-13
buildFileIndex: byExact double-indexing causes both phantom keys and unresolvable extension-less files


JordanCoin:
Pre-existing on `main`. One three-line block produces two opposite defects, and it is the shared root cause behind findings in #125, #126 and #127.

## The code

`scanner/filegraph.go:248-250`:

```go
idx.byExact[path] = append(idx.byExact[path], path)
noExt := strings.TrimSuffix(path, filepath.Ext(path))
idx.byExact[noExt] = append(idx.byExact[noExt], path)
```

## Defect A — phantom keys (causes false positives)

A file named `src/bindings.rs.in` has `Ext == ".in"`, so `noExt == "src/bindings.rs"`. The index now contains a key for a path **that may not exist on disk**, with exactly one entry.

Any resolver using the common idiom `if len(idx.byExact[target]) != 1 { return "" }; return target` will therefore return a path to a nonexistent file. This is the mechanism behind the fabricated edge in #125:

```
$ codemap --json --importers src/bindings.rs
{"file":"src/bindings.rs","importers":["src/main.rs"],"importer_count":1}
$ ls src/bindings.rs
No such file or directory
```

`foo.rs.in` beside a build-time-generated, gitignored `foo.rs` is a normal codegen layout, so this is reachable in ordinary repos.

## Defect B — extension-less files are permanently unresolvable (causes false negatives)

When a path has no extension, `filepath.Ext` returns `""` and `noExt == path`, so the **same path is appended twice under the same key**:

| path | ext | noExt | noExt == path |
|---|---|---|---|
| `app/VERSION` | `""` | `app/VERSION` | **true** |
| `Makefile` | `""` | `Makefile` | **true** |
| `app/schema.proto` | `".proto"` | `app/schema` | false |

Every extension-less file has `len(byExact[p]) == 2`, so the same `len(...) != 1` idiom rejects it forever. #127 found this concretely: `cargo` proves `app/VERSION` is a declared build-script input, and codemap reports `No files import app/VERSION`. It hits `VERSION`, `Makefile`, `Dockerfile`, `LICENSE`, and extension-less config — all common `rerun-if-changed` targets.

## Why fixing it centrally is worth more than three local patches

Three open PRs touch this idiom:

- **#125** `resolveRustInclude` — hits defect A, produces a fabricated edge (blocker).
[...]

PR #142, opened 2026-08-22
fix(scanner): Track Rust embedded file dependencies

reneleonhardt:
## What does this PR do?

Record dependencies from literal Rust `include_str!` and `include_bytes!` calls. Resolve relative paths only when the exact target is uniquely indexed in the repository; leave dynamic, missing, external, excluded, ambiguous, self, and `include!` targets unresolved.

## Type of change

- [x] Bug fix

## Verification

- Focused tests cover literal and raw strings, dynamic, missing, external, excluded, ambiguous, and self targets.
- `go vet ./...`
- `go test ./...`

### DIFF
--- a/scanner/astgrep.go
+++ b/scanner/astgrep.go
@@ -479,7 +479,7 @@ func (s *AstGrepScanner) scanDirectory(parent context.Context, root string) ([]F
 				}
 			}
 			if path != "" {
-				if m.RuleID != "rust-path-imports" && m.RuleID != "rust-askama-template-imports" && m.RuleID != "rust-cargo-rerun-imports" {
+				if m.RuleID != "rust-path-imports" && m.RuleID != "rust-askama-template-imports" && m.RuleID != "rust-embedded-file-imports" && m.RuleID != "rust-cargo-rerun-imports" {
 					fileMap[relPath].Imports = append(fileMap[relPath].Imports, path)
 				}
 				fileMap[relPath].References = append(fileMap[relPath].References, ImportReference{
--- a/scanner/astgrep.go
+++ b/scanner/astgrep.go
@@ -421,7 +421,7 @@ func (s *AstGrepScanner) scanDirectory(parent context.Context, root string) ([]F
 			}
 		}
 
-		if m.RuleID == "rust-mod-imports" || m.RuleID == "rust-path-module-imports" || m.RuleID == "rust-path-imports" || m.RuleID == "rust-use-imports" || m.RuleID == "rust-askama-template-imports" || m.RuleID == "rust-include-imports" || m.RuleID == "rust-cargo-rerun-imports" {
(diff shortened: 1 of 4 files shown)


JordanCoin:
Reviewed in depth (adversarial pass with end-to-end probes on the PR head). The rule scoping is solid — `concat!`/`env!` args correctly yield nothing, raw strings parse right, traversal/self-includes are guarded, suite is green. Two blocking items before this lands, both small:

## 1. BLOCKING — phantom edge: the resolver's uniqueness check accepts extension-stripped index keys

`resolveRustEmbeddedFile` uses a bare `len(idx.byExact[target]) != 1`, but `buildFileIndexContext` also indexes every file under its extension-stripped key. Demonstrated end-to-end: a repo with `src/shader.wgsl.tmpl` (checked in) and no `src/shader.wgsl` — `include_str!("shader.wgsl")` resolves and the graph gains `src/lib.rs → src/shader.wgsl`, **a file that doesn't exist on disk**. Same pattern for `.license`, `.in`, `.orig` siblings.

Fix is the idiom your own sibling functions already use (`resolveRustInclude` documents this exact hazard): count entries **equal to** `target` and require exactly 1. That also fixes the inverse miss (real `data.txt` unresolvable because `data.txt.license` exists). Please also restore the `TestResolveRustIncludeRequiresRealIndexedFile` regression test this PR removed — it was the guard for precisely this class — and add the embedded-file analog.

## 2. BLOCKING — `rust-embedded-file-imports` missing from the Imports-append exclusion list

`scanner/astgrep.go` excludes `rust-path-imports`, `rust-askama-template-imports`, and `rust-cargo-rerun-imports` from the raw `imports` array, but not the new kind — so asset paths like `data.txt` land in the versioned payload and feed `--diff`'s basename fuzzy matching (a change to unrelated `docs/data.md` would falsely count the including file as impacted). Same invariant #127 settled. Add the exclusion and drop the two asset entries from the test's `imports` expectation (the test currently asserts the leak).

## Nits (non-blocking)

- Extensionless targets (`VERSION`, `LICENSE`) can never resolve due to the pre-existing #130 index bug — worth a code comment since `include_str!` hits them often; the fix belongs in #130, not here.
- The scan test constructs two scanners; the second (which actually scans) is never closed and shadows the package name — looks like a rebase artifact.

## Rebase note

#143 and #144 just merged; this branch auto-merges cleanly with the new main (verified with real three-way merges — your `astgrep.go`/`rustgraph.go` hunks don't overlap #143's), so a plain rebase or merge of main should be conflict-free. Re-run the suite after; with the two fixes above this is ready to land.

---

reneleonhardt:
### Complied

- Require an exact indexed path, preventing phantom embedded-file edges.
- Restored the real-file include regression and added the embedded-file equivalent.
- Excluded `rust-embedded-file-imports` from raw `imports`.
- Updated expectations and closed the actual scanner instance.

### Not fully complied

- Extensionless targets remain unresolved due to the pre-existing shared-index issue; documented and left for #130.

Issue #155, opened 2026-08-24
Design: the watch daemon has no owner, no serialization, and no reaper


JordanCoin:
**Version:** 4.4.2 (Homebrew) · macOS
**Related:** the "`hook session-start` silently spawns a daemon" bug report — that one is about it being *undocumented*. This one is about whether the daemon should work this way at all.

## What I found

A single project directory, observed live:

```
PID    ELAPSED       COMMAND
44058  5d 04:22:41   codemap watch daemon <project>
81231     06:54:57   codemap watch daemon <project>
81271     06:54:56   codemap watch daemon <project>
81298     06:54:55   codemap watch daemon <project>
```

Four daemons watching the same directory. Three of them started **within one second of each other**. The fourth had been running for five days, long after whatever session started it had ended.

That's two separate defects wearing one costume.

## Defect 1 — the "already running?" check is not a guard

Three spawns at `:55`, `:56`, `:57` means the check is read-then-act with a window in between. Any two invocations that land inside that window both conclude nothing is running and both spawn. Editors, shells, and agent tooling all tend to fire session hooks at once, which is exactly the access pattern that loses this race.

A liveness check is not mutual exclusion. This needs an atomic claim — `O_EXCL` lock file holding the pid, or a unix socket bind, or flock — where losing the claim means exiting rather than proceeding.

## Defect 2 — nothing owns the daemon, so nothing reaps it

The five-day-old process is the interesting one. It outlived its session by days, still watching, still holding state, invisible to the user who caused it. There is no parent that dies and takes it with it, no TTL, no "is my originating session still alive" heartbeat, and no documented way to stop one — `--help` lists `--watch` to start but nothing to stop.

Over weeks this accumulates silently. I only noticed because I went looking for something else.

## The design question underneath

The part I'd actually reconsider is `hook session-start` spawning it at all.

A watcher you opt into with `--watch` and can see is a normal tool. A watcher that appears because you asked for a *printout* — under a subcommand whose help text is "Show project context" — and then outlives the thing that asked, is the shape that produced four copies. The user never formed an intent to run a background process, so they never form an intent to stop one.

## Is the daemon buying anything?

Worth asking honestly, because the costs above are real and the benefit may not be. Measured cold, no daemon, on a ~140-file Go repo and a ~290-file Swift repo:

| command | runtime |
|---|---|
| `handoff` | 0.59s / 1.00s |
| `--importers <file>` | 0.30s / 0.52s |
[...]

Issue #140, opened 2026-08-15
watch: writeState discards the os.WriteFile error, and get_file_context reports confident zeros for files that do not exist


JordanCoin:
Two small silent-failure bugs, same shape: an operation that cannot succeed reports as though it did.

## 1. `writeState` throws away its write error

`watch/events.go:675`:

```go
stateFile := filepath.Join(d.root, ".codemap", "state.json")
os.WriteFile(stateFile, data, 0644)      // return value discarded
}
```

If that write fails — read-only `.codemap`, disk full, permissions, a transient error — the daemon proceeds as though state was persisted. Every hook that later calls `ReadState` gets the stale file or nothing, and silently falls back to a scan.

`writeState` is called from the debounce loop and on every control event, so a persistent failure repeats indefinitely with no signal anywhere. The daemon believes it is publishing state; nothing is.

Minimal fix: check the error and surface it once (log under `d.verbose`, or record it so `codemap watch status` can report a degraded daemon). Failing loudly once is better than failing silently forever.

## 2. `get_file_context` answers confidently about files that do not exist

Driving the MCP server against a path that is not in the repo:

```
$ get_file_context {"file": "totally/made/up.go"}

=== File Context: totally/made/up.go ===

IMPORTS: none (leaf file)

IMPORTED BY: none (entry point or unused)

CONNECTED: 0 files in dependency graph
```

No error, no caveat. The parenthetical explanations make it worse than bare zeros: **"leaf file"** and **"entry point or unused"** are assertions about a file that was never scanned and does not exist. An agent that mistypes a path, or asks about a file from a stale mental model, is told the file is a harmless leaf.

Minimal fix: check membership in the scanned inventory first and return "not found in the scanned file set" — distinguishing *unknown* from *known-and-empty*. The index already has the file list, so this is a lookup, not new machinery.

## Why these are worth filing despite being small

Both are the house failure mode: a confident answer where the honest one is "I don't know" or "that failed." (2) is a near-duplicate of the `--importers` problem in #138 on a different surface, which suggests the fix belongs at the layer where an answer about a file is assembled, not in each command.

Related: #138 (same class, `--importers`), #137 (the disclosure channel these should use), #111.

PR #158, opened 2026-08-27
fix(watch): Scope runtime state per project

reneleonhardt:
## What does this PR do?

Scopes mutable watcher state to the canonical project root and hardens watcher lifecycle handling:

- Stores pid, state, events, and handoff files in a per-project runtime namespace.
- Validates linked-worktree and runtime roots before reading or writing state.
- Waits for detached watcher readiness and preserves bounded cancellation behavior.
- Keeps setup configuration in the configured `.codemap` directory.

## Why it matters

Projects sharing a setup root could collide on daemon state or stop one another’s watcher. Invalid linked-worktree metadata could also produce misleading success or raw git errors. This keeps ownership and lifecycle outcomes explicit.

## CLI / MCP surface

No new commands, arguments, or MCP tools. Existing watch, hook, and handoff flows use project-scoped runtime state.

## Security

Runtime markers are accepted only for the canonical owning project and verified process. Invalid roots and foreign ownership fail closed.


JordanCoin:
Reviewed in depth with live daemon testing (concurrent starts, `kill -9` recovery, pid-reuse attacks, symlink/worktree keying — details at the bottom). The lifecycle core is genuinely solid, and the ownership gate on `terminateDaemon` fixes a real pre-existing hole where Unix would SIGTERM an unverified pid. One blocking item and one worth fixing in the same push:

## 1. BLOCKING — legacy-daemon detection misses the immediately-prior `--setup-root` layout, so upgrading duplicates daemons

`ResolveActiveRuntime`'s legacy fallback (`watch/state.go:152`, via `internal/projectpath/runtime.go:26-45`) only covers the pre-#122 flat layout (`<setup>/.codemap/watch.pid`). It does not cover what main used until this PR for `--setup-root` projects: `<setup>/.codemap/projects/<key>/watch.pid`. Reproduced end-to-end: daemon started with a main-built binary under `--setup-root`, then `watch status` with this PR's binary → "not running"; `watch start` → a **second live daemon** watching the same directory (both pids alive, old and new pid files coexisting). That silently double-watches exactly the shared-setup-root scenario this PR is titled for. Fix: add `<setup>/.codemap/projects/<ProjectKey>` as a legacy candidate in `ResolveActiveRuntime`'s list — and a test for it; `TestResolveActiveRuntimeUsesOnlyExactlyOwnedLiveLegacyDaemon` currently only exercises the flat layout.

## 2. Fix in the same push — `Release()` before `Pid` is read

`main.go:809-816`: `os.Process.Release()` sets `Pid = -1` on Linux, and the code releases before reading. Every successful start prints "Watch daemon started (pid -1)" (confirmed), and on the readiness-timeout branch `RemoveProcessPID(absRoot, cmd.Process.Pid)` receives `-1`, so `removePIDIfMatches` no-ops and the just-published pid file is left behind (the next start recovers it via the stale path, but the designed cleanup never runs). Capture `pid := cmd.Process.Pid` before `Release()`, or defer the release past both uses.

## Non-blocking

- `RuntimeCodemapDir` (`internal/projectpath/path.go:172-184`) keeps a silent fail-open fallback on `SelectRuntime` error. Every current production caller pre-validates, so it's latent — but it undercuts the PR's own fail-closed invariant for the next caller. A checked counterpart (essentially `CheckedRuntimeCodemapDir`) routed through the watch-state call sites would close it.
- `mcp/main.go`: `handleStartWatch` no longer populates the in-process `watchers` map (good — delegating to the subprocess gives MCP the same lock/readiness/ownership guarantees), which leaves the `daemon != nil` branches in `handleStopWatch`/`handleGetActivity` production-unreachable. Fine as a follow-up cleanup.

## What held up under attack

Two-project isolation with `kill -9` + restart (stale `.pid.stale` evidence preserved); pid-reuse — a stale pid file pointing at an unrelated live process gets `ErrForeignDaemonPID` and the process is never signaled; concurrent `watch start` for one project serialized by the transition lock (one daemon, clean "already running" for the loser, a dozen trials); symlink vs real path, trailing slash, and linked worktree vs primary all key identically; plain non-git directories unchanged; invalid `.codemap` state fails fast (~6ms) with a bounded error before forking. `go vet` clean (including a windows/amd64 cross-compile), full suite green apart from the three known root-environment tests, `go test ./watch -count=3` and `-race -count=2` stable.

## Sequencing / issue notes

No conflicts with main or the open #159 (verified via merge-tree; the only shared file is `main.go` in disjoint regions — either merge order is safe). On the open issues: this meaningfully advances #155 (serialization + ownership; the reaper gap remains — a daemon whose project dir is deleted still becomes unstoppable, same as main), and doesn't touch #140/#151/#152 despite being adjacent — worth "Relates to #155" in the merge, nothing stronger.

With items 1–2 fixed, this is ready to land.

---

reneleonhardt:
## Complied

- Added migration detection and regression coverage for the immediately-prior project-scoped `--setup-root` daemon layout.
- Captured the daemon PID before release and reused it for readiness cleanup and startup output.

JordanCoin:
Verified and merged. Reproduced the migration scenario live against the new head: a daemon started by a main-built binary under `--setup-root` in the prior layout is now seen by the new binary's `status`, `start` refuses to duplicate ("already running", one process), and `stop` cleanly terminates it. The pid capture fix checks out in the diff and both new test assertions guard it. Full `go vet` + `go test ./...` clean against main. Solid work on the lifecycle hardening — the ownership gate before signaling closes a real pre-existing hole.

---

PR #161, opened 2026-08-29
fix(importers): Canonicalize absolute importer paths

reneleonhardt:
## What does this PR do?

Makes absolute `--importers` queries use the same canonical project and file roots:

- Infers the nearest Git project for an absolute importer launched outside the project.
- Canonicalizes the importer in both explicit-root and inferred-root paths.
- Preserves explicit project-root precedence.
- Adds end-to-end coverage for canonicalized reports and root precedence.

## Why it matters

On macOS, `/tmp` and `/private/tmp` can name the same file. Mismatched spellings produced incorrect relative paths and empty importer results.

## CLI / MCP surface

No new commands, arguments, or MCP tools. Existing `--importers` queries now work consistently with absolute paths.

## Verification

The repository linter reports no new findings for this branch beyond its existing baseline.

### DIFF
--- a/blast_radius.go
+++ b/blast_radius.go
@@ -1412,7 +1412,13 @@ func renderCoverage(w io.Writer, status string, notes []string) {
 }
 
 func buildImportersReportFromGraph(root, file string, fg *scanner.FileGraph) scanner.ImportersReport {
+	if canonical, err := filepath.EvalSymlinks(root); err == nil {
+		root = canonical
+	}
 	if filepath.IsAbs(file) {
+		if canonical, err := filepath.EvalSymlinks(file); err == nil {
+			file = canonical
+		}
 		if rel, err := filepath.Rel(root, file); err == nil {
 			file = rel
 		}
--- a/main.go
+++ b/main.go
@@ -58,7 +58,7 @@ type watchReadiness struct {
 }
 
 func main() {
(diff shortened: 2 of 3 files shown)


JordanCoin:
Verified and merged. Canonicalization holds up under the attacks that matter: cross-repo symlink aliasing, nested repo-in-repo, a submodule, and a linked worktree (`.git` as a file) all resolve to the file's true identity — I couldn't get it to map one file onto another's identity. Dangling symlinks and no-git-ancestor paths fail closed with bounded errors rather than a confident empty answer. Relative-path `--importers` output is byte-identical to main across a dozen real files, and the new tests are load-bearing (reverting the fix fails three of the four). Clean merge, suite at baseline.

One adjacent gap worth a follow-up, **pre-existing and not from this PR** (confirmed byte-identical on main, so nothing here regressed): in `buildImportersReportFromGraph` the `filepath.Rel(root, file)` result isn't checked for staying inside `root`. With an explicit root and an absolute `--importers` path outside it, `Rel` happily walks up and the tool prints `No files import ../../home/user/...` with exit 0 instead of erroring — a confident empty answer in exactly the shape this project treats as the cardinal sin. Guarding on `strings.HasPrefix(rel, "..") || filepath.IsAbs(rel)` would close it; happy to look at that separately whenever you want it picked up.

---

PR #163, opened 2026-08-29
feat(topology): Map SwiftPM targets

reneleonhardt:
## What does this PR do?

Adds SwiftPM target topology discovery from `Package.swift`:

- Parses package products, targets, target paths, and declared target dependencies.
- Resolves target paths conservatively into the typed topology graph.
- Keeps malformed or ambiguous declarations bounded and unresolved.

## Why it matters

Swift packages define module boundaries and dependencies in `Package.swift`; directory-only topology misses that structure.

## CLI / MCP surface

No new commands, arguments, or MCP tools. Existing topology output includes SwiftPM package and target nodes.

## Security

`Package.swift` is parsed as source text and never executed.

## Review follow-ups included

- Ignores package-shaped syntax inside normal, raw, and multiline Swift strings.
- Rejects computed or interpolated target values while accepting escaped delimiters and literal string forms.

## Verification

### DIFF
--- a/topology/swiftpm_parser.go
+++ b/topology/swiftpm_parser.go
@@ -331,9 +331,10 @@ func swiftBalancedBody(text string, start int, open, close byte) (string, int, b
 			continue
 		}
 		char := text[index]
-		if char == open {
+		switch char {
+		case open:
 			depth++
-		} else if char == close {
+		case close:
 			depth--
 			if depth == 0 {
 				return text[start+1 : index], index + 1, true
@@ -438,12 +439,19 @@ func swiftStringInfo(text string, start int) (valueStart, valueEnd, end int, ok 
 		valueStart = quote + 3
 		close = `"""` + strings.Repeat("#", hashes)
 	}
-	closeOffset := strings.Index(text[valueStart:], close)
-	if closeOffset < 0 {
-		return valueStart, len(text), len(text), false
+	for index := valueStart; index < len(text); index++ {
+		if text[index] == '\\' {
+			index++
+			for index < len(text) && text[index] == '#' {
+				index++
+			}
+			continue
+		}
+		if strings.HasPrefix(text[index:], close) {
+			return valueStart, index, index + len(close), true
+		}
 	}
-	valueEnd = valueStart + closeOffset
-	return valueStart, valueEnd, valueEnd + len(close), true
+	return valueStart, len(text), len(text), false
 }
 
 func swiftStringEnd(text string, start int) (int, bool) {
@@ -493,9 +501,10 @@ func swiftHasInterpolation(value string) bool {
 
 func swiftPMConventionalRoot(packageRoot, call, name string) string {
 	base := "Sources"
-	if call == "testTarget" {
+	switch call {
+	case "testTarget":
 		base = "Tests"
-	} else if call == "plugin" {
+	case "plugin":
 		base = "Plugins"
 	}
 	return filepath.Clean(filepath.Join(packageRoot, base, name))
--- a/topology/swiftpm_parser.go
+++ b/topology/swiftpm_parser.go
@@ -1,0 +1,482 @@
+package topology
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+)
+
+type swiftPMManifest struct {
+	manifest string
+	root     string
+	targets  []swiftPMTarget
+	products []swiftPMProduct
+	issues   []Issue
+}
+
+type swiftPMTarget struct {
+	name         string
+	kind         NodeKind
+	root         string
+	sourceRoots  []string
+	testRoots    []string
+	memberRoots  []string
+	excludes     []string
+	dependencies []swiftPMDependency
+}
+
+type swiftPMProduct struct {
+	name    string
+	targets []string
+}
+
+type swiftPMDependency struct {
+	kind        string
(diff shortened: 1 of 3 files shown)


JordanCoin:
Reviewed against the #160 provider contract with a hostile-fixture matrix run end-to-end through `--topology --json`. The dynamic-construct rejection here is genuinely strong — see the negative evidence at the bottom, it's a lot. One critical blocker and one I'd want fixed alongside it.

## 1. BLOCKING — nested block comments leak package-shaped text into a real node

`stripSwiftComments` (`topology/swiftpm_parser.go:381-426`) tracks block comments with a single boolean, no depth counter. Swift block comments **nest**, so the scanner closes the outer comment at the first `*/`. Reproduced end-to-end on a built binary:

```swift
targets: [
    .target(name: "Real"),
    /* docs: /* inner */ .target(name: "Ghost"), */
    .target(name: "Real2"),
]
```
→
```json
"swiftpm:Package.swift:Ghost": { "kind": "swift-target", "name": "Ghost", "root": "Sources/Ghost", ... }
```

Two harms in one manifest: a **fabricated `Ghost` node** that exists only inside a comment, and the **real `Real2` target disappears** (corrupted by the same leak). Coverage does go `partial` — but only because `Real2` failed to parse; the `Ghost` node itself carries no Issue and no unresolved/ambiguous marker, so a consumer cannot tell it from a genuine one. This is the same class as the CUE lexer in #146, and the fix is the same shape: track nesting depth in `stripSwiftComments` (increment on `/*`, decrement on `*/`, exit only at depth 0).

## 2. Please fix alongside — raw-string escape handling erases whole-file coverage on valid syntax

`swiftStringInfo` (`topology/swiftpm_parser.go:442-451`) treats a bare `\` as an escape introducer in raw strings, where Swift only treats `\#` (with matching hash count) as special. So `path: #"foo\"#` — a perfectly ordinary raw string — makes the scanner skip the real closing delimiter, run off the end, and reject the entire `Package(...)` call as non-literal: **0 targets**, `computed-swiftpm-package`. It fails closed rather than fabricating, which is the right direction, but one legal backslash silently erases every target and product in the file. Fix: in raw strings, treat `\` as special only when followed by exactly `hashes` `#` then `(`.

## Non-blocking

- Two targets declaring the same `path:` are both assigned the full overlapping file set with `Coverage.Status` still `Complete` and no ambiguity Issue (`swiftpm_provider.go:197-228`). Invalid SwiftPM anyway, but double-ownership should be flagged rather than silent.
- `path:` values with `..` or absolute paths are joined and cleaned without rejection or an Issue (`swiftpm_parser.go:106-114`), so a node's root/membership can point outside the package tree with no signal. Metadata-only, but worth an Issue given the conservative framing.

## What held up (verified negative evidence)

Single-level comments, well-formed normal/multiline/raw strings at any hash count with package-shaped text inside, escaped quotes, raw multiline `#"""..."""#`, targets built via loop or `.map`, `.target(name: someVariable)`, string concatenation, conditional `.product(condition: .when(...))` (flagged conditional, no edge fabricated), `#if os(...)` blocks (dropped with an Issue, coverage partial), targets from a helper function, and a comment inside a target's argument list — all correctly refused or bounded, never a fabricated node. A 20k-target/1.4MB manifest parsed in ~264ms, iterative throughout, so no blowup or stack risk. `Package.swift` is genuinely never executed (pure byte scanning; verified by reading — swift isn't installed here).

## Sequencing

Clean against both #162 and #165 — zero overlapping files, each provider self-registers via `init()`, so the trio is a pure mechanical union. `go vet` clean; full suite matches the main baseline exactly.

---

reneleonhardt:
### Complied

- Nested Swift block comments now track depth.
- Raw strings recognize only exact-hash interpolation escapes.
- Unsafe target paths are rejected.
- Duplicate target roots produce ambiguity issues and partial coverage.
- Added focused hostile regressions.

### Additional changes

- Corrected package-root containment for `"."` roots.

JordanCoin:
Verified and merged. Re-ran both original reproductions against `0082f6b`:

- **nested block comment** — `/* docs: /* inner */ .target(name: "Ghost"), */` no longer fabricates `Ghost`, and the real `Real2` target that the leak was swallowing is recovered. Nodes are exactly `Real` + `Real2`.
- **raw string `#"foo\"#`** — was wiping the whole file to 0 targets; now both `A` and `B` parse normally.

The extras landed too: duplicate target roots raise `ambiguous-swiftpm-target-path` with partial coverage, and unsafe `path:` values are rejected. `go vet` clean, suite at baseline.

One thing I noticed while verifying, **not caused by this PR** and not a reason to hold it — I'm flagging it here because it touches how your coverage reads: a clean single-target package reports `Coverage: partial` with zero issues. It comes from `BuildGraph` in `topology/graph.go`:

```go
if fragment.Coverage.Status != CoverageComplete { complete = false }
```

The JVM provider (correctly) returns `CoverageUnavailable` when a repo has no Maven/Gradle/sbt manifests, and that "not applicable here" gets treated the same as "couldn't fully understand this", so any non-JVM repo lands in the `default:` → `partial` branch with nothing to explain it. Confirmed on current main independent of this PR: a minimal PEP 621 Python project reports `partial`, a Maven-only repo reports `complete`. I'll follow up on that separately.

---

PR #165, opened 2026-08-29
feat(topology): Map JVM projects

reneleonhardt:
## What does this PR do?

Adds JVM project topology for Maven, Gradle, and sbt repositories. It maps local projects, build boundaries, literal local-project dependencies, and conventional or declared Java, Kotlin, and Scala source roots. It supports legacy and current manifest forms where the syntax is safely recognizable and reports partial coverage for dynamic or ambiguous declarations.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

### DIFF
--- a/blast_radius.go
+++ b/blast_radius.go
@@ -520,7 +520,11 @@ func buildBlastRadiusBundle(absRoot, ref string, limits blastRadiusLimits) (blas
 		// Analyze the FULL changed set; capping only limits what is displayed,
 		// it must not shrink blast-radius analysis or the summary stats.
 		for _, file := range changedFiles {
-			allReports = append(allReports, buildImportersReportFromGraph(absRoot, file.Path, fg))
+			report, reportErr := buildImportersReportFromGraph(absRoot, file.Path, fg)
+			if reportErr != nil {
+				return blastRadiusBundle{}, reportErr
+			}
+			allReports = append(allReports, report)
 		}
 		shownReports = allReports
 		if len(diffCapped.Files) < len(allReports) {
@@ -1411,18 +1415,25 @@ func renderCoverage(w io.Writer, status string, notes []string) {
 	fmt.Fprintf(w, "Coverage: %s — %s\n", status, strings.Join(notes, "; "))
 }
 
-func buildImportersReportFromGraph(root, file string, fg *scanner.FileGraph) scanner.ImportersReport {
+func buildImportersReportFromGraph(root, file string, fg *scanner.FileGraph) (scanner.ImportersReport, error) {
 	if canonical, err := filepath.EvalSymlinks(root); err == nil {
 		root = canonical
 	}
-	if filepath.IsAbs(file) {
-		if canonical, err := filepath.EvalSymlinks(file); err == nil {
-			file = canonical
-		}
-		if rel, err := filepath.Rel(root, file); err == nil {
-			file = rel
-		}
+	filePath := file
+	if !filepath.IsAbs(filePath) {
+		filePath = filepath.Join(root, filepath.FromSlash(filePath))
 	}
+	if canonical, err := filepath.EvalSymlinks(filePath); err == nil {
+		filePath = canonical
+	}
+	rel, err := filepath.Rel(root, filePath)
+	if err != nil {
+		return scanner.ImportersReport{}, fmt.Errorf("resolve file %q relative to project root %q: %w", file, root, err)
+	}
+	if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) || filepath.IsAbs(rel) {
+		return scanner.ImportersReport{}, fmt.Errorf("file %q is outside project root %q", file, root)
+	}
+	file = rel
 	file = filepath.ToSlash(file)
 
 	// Preserve scan order (do not sort): this helper also backs the pre-existing
@@ -1447,7 +1458,7 @@ func buildImportersReportFromGraph(root, file string, fg *scanner.FileGraph) sca
 			report.HubImports = append(report.HubImports, imp)
 		}
 	}
-	return report
+	return report, nil
 }
 
 func printAstGrepInstallHint(w io.Writer, err error) {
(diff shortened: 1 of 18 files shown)


JordanCoin:
Reviewed against the #160 provider contract, with every finding below reproduced end-to-end on a binary built from this head (`--topology --json`). **Maven and sbt are in good shape; Gradle is the weak link** — it's missing the block-comment state machine that `sbt.go` in this same PR gets right, and its literal extractor can't tell a literal from interpolated text. Four Gradle issues, three of which produce confidently-wrong output under `"status": "complete"`.

## 1. BLOCKING — `/* */` block comments aren't stripped; commented-out includes become real nodes

`stripGradleLineComment` (`topology/gradle.go:499-522`) only handles `//`; there's no multi-line state, unlike `stripSBTComments` (`sbt.go:293-332`) which threads `inBlockComment` correctly.

```groovy
rootProject.name = "commentattack"
include(":app")
/*
include ':experimental'
*/
```
→ node `gradle:settings.gradle::experimental` with `root: "experimental"` and source roots, under `"coverage": {"status": "complete"}`. Same class as the CUE lexer in #146. Fix: give the Gradle scanner the block-comment state `stripSBTComments` already has, threaded across lines in both `parseGradleSettings` and `parseGradleBuildFile`.

## 2. BLOCKING — interpolated strings are accepted as literal project paths

`gradleStringLiteralPattern` (`gradle.go:50`) accepts anything between quotes, including `$`/`${}` content.

```groovy
include(":${name}")     →  node  gradle:settings.gradle::${name}
include(":$m")          →  node  gradle:settings.gradle::$m
```
both with `"status": "complete"`. (Note: the *loop* form `for (m in modules) { include(":$m") }` doesn't produce a fake node — it's silently dropped instead, which is finding 4's problem. The direct single-line form is what fabricates.) Fix: reject any literal containing `$` before accepting it as a project path, falling back to the existing `dynamic-gradle-*` issue codes.

## 3. BLOCKING — `subprojects {}` / `allprojects {}` content is attributed to the declaring file's own project

The line scanner has no block-context tracking, so a dependency declared for all subprojects is recorded as the *root* project's own:

```groovy
// root build.gradle
subprojects {
    dependencies {
        implementation project(':common')
    }
}
```
→ edge `gradle:settings.gradle:: → gradle:settings.gradle::common`, `"status": "complete"`. The root has no such dependency in reality, and the edge is missing from the subprojects that do. Worth noting the single-line variant (`dependencies { implementation project(':common') }` on one line) *is* correctly refused with `dynamic-gradle-dependency` — so the guard exists, it just doesn't survive the multi-line form, which is the common style. Fix: detect `subprojects`/`allprojects` blocks and either attribute to the siblings or back off to a partial-coverage issue.

## 4. Coverage honesty — multi-line `include(...)` silently dropped while claiming complete

`gradleIncludeCallPattern` is anchored to a single line, so:
```groovy
include(
    ":multiline-a",
    ":multiline-b"
)
```
yields zero nodes for those projects and **no issue** — `"status": "complete"`. A missed node is fine by house rules; claiming full confidence while missing it isn't. Same applies to sbt projects built programmatically (`Seq(...).map(n => Project(n, ...))`) — invisible with no `dynamic-sbt-*` issue, which doesn't match the PR's "reports partial coverage for dynamic or ambiguous declarations" claim.

## Non-blocking

- Maven `<module>${prop}</module>` isn't run through `resolveMavenValue` before the directory join (`maven.go:186-188`), so a property-interpolated module with a real dir/pom still reports `unresolved-maven-module: declares missing local module "${core.module}"` and drops the parent→child boundary edge. Fails closed correctly (partial, no false edge) — just an accuracy gap, and the resolver you need is already used elsewhere in that file. Same for `<relativePath>`.
- `internal/projectpath.CanonicalPath` (new here, bundled to fix watch path-identity flakes) now calls `EvalSymlinks` per directory and per scanned path in `ScanFiles`, so it runs for every provider's inventory scan, not just JVM. No measurable regression on a 2,400-file tmpfs tree, but tmpfs hides stat cost — worth one benchmark on a large real-disk repo before merge.

## What held up

Maven is disciplined: XXE and billion-laughs both fail the document cleanly (Go's `encoding/xml` has no DTD entity support and nothing enables it) → `unavailable` + `provider-failed`, no crash or hang; profile-controlled `<modules>` correctly flags `maven-profile-topology` and never asserts the profile-only module; missing module dirs degrade to partial. Gradle type-safe accessors (`projects.myLib` → `:my-lib`) resolve correctly; bare-variable `include(projectName)` and a `projectDir` escaping the build root are both caught by your own tests. sbt's block comments are handled correctly and its literal `dependsOn`/`aggregate` resolution is sound. One provider (`Name() == "jvm"`) covers all three build systems, so no duplicate-registration risk.

## Sequencing

Clean union with #162 and #163 — zero overlapping files, distinct provider names. `go vet` clean; suite failure set identical to main.

---

JordanCoin:
Verified and merged. Re-ran all four original reproductions against `c10cac2`:

| case | before | now |
|---|---|---|
| commented-out `include ':experimental'` | phantom node, `complete` | gone — only root + app |
| multi-line `subprojects { dependencies { … } }` | false `root → common` edge, `complete` | no edge, `partial` + `dynamic-gradle-scope` |
| `include(":${name}")` | literal `${name}` node, `complete` | no node, `partial` + `dynamic-gradle-include` |
| multi-line `include(...)` | silently dropped, `complete` | dropped but `partial` + `dynamic-gradle-include` |

Also confirmed the two non-blocking items landed: `<module>${core.module}</module>` with a real `core/pom.xml` now resolves the module and the parent→child boundary edge (was `unresolved-maven-module`), and the out-of-root `--importers` guard I mentioned on #161 now returns a bounded error with exit 1 instead of a confident `No files import ../../…` — nice pickup, that one wasn't in your review list.

`TestDebounce` still green here (worth noting since #164 currently regresses it), `go vet` clean, full suite at the main baseline.

---

PR #166, opened 2026-09-01
fix(topology): Don't let non-applicable providers force partial coverage

JordanCoin:
## What does this PR do?

Fixes a coverage-reporting defect I let through while merging #165: **every repo without JVM or SwiftPM manifests reports `partial` coverage with no issue explaining why.**

`MergeFragments` computed completeness as:

```go
if fragment.Coverage.Status != CoverageComplete { complete = false }
```

The JVM and SwiftPM providers correctly return `Unavailable` with no nodes, edges or issues when their manifests are absent — meaning *"not applicable to this repo"*. The aggregation read that as *"could not fully understand this repo"* and downgraded the whole graph. Measured on current main:

| repo | before | after |
|---|---|---|
| clean PEP 621 Python project | `partial` (no issues) | `complete` |
| Maven-only | `complete` | `complete` |
| SwiftPM-only | `partial` (no issues) | `complete` |
| no manifests at all | `unavailable` | `unavailable` |
| Python with a real layout issue | `partial` + issue | `partial` + issue |
| polyglot (Python + Maven) | `partial` (no issues) | `complete` |

Coverage errs safe today rather than overclaiming, so nothing was ever reported as more certain than it is — but `partial` firing on nearly every repo with nothing to point at makes the signal unusable, which defeats the purpose of the contract #160 established.

## The fix

A fragment that declares `Unavailable` **and** carries no nodes, edges, issues or members is treated as not-applicable: neutral for completeness. Two properties are deliberately preserved:

- An otherwise-empty graph still reports `unavailable` rather than "complete knowledge of nothing" — that's what the `sawNotApplicable` flag protects.
- A provider that declares `Complete` with no nodes ("I looked, there is genuinely nothing here") still counts as complete and stays cacheable, and a real `provider-failed` fragment carries an issue so it still degrades coverage.

That last distinction came out of `TestBuildGraphUsesProjectRuntimeCache`: my first attempt made *any* empty graph unavailable, which made legitimately-empty results uncacheable and rebuilt the graph on every call. The narrower rule keeps that test passing on its own terms.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other (describe below)

## Checklist

## Additional notes

`TestMergeFragmentsIgnoresNonApplicableProviders` covers all three branches and is load-bearing — against unmodified `graph.go` it fails with `coverage = "partial", want "complete"`. Verified end-to-end with a built binary across the six repo shapes in the table above. `go vet ./...` clean; full `go test ./...` at the main baseline (only the three known root-environment permission failures).

The affected code is @reneleonhardt's from #160 and #165 — flagging since this adjusts the aggregation semantics they designed, and I'm happy to defer if you'd rather shape it differently.

---

### DIFF
--- a/topology/graph.go
+++ b/topology/graph.go
@@ -34,7 +34,19 @@ func MergeFragments(root string, fragments []Fragment) *Graph {
 	var issues []Issue
 	invalidIDs := make(map[ID]bool)
 	complete := true
+	sawNotApplicable := false
 	for _, fragment := range fragments {
+		// A provider that found none of its manifests reports Unavailable with
+		// nothing to say: "not applicable here", not "incomplete". Such a
+		// fragment must not drag down a graph that other providers answered
+		// fully, but it does mean an empty graph is unavailable rather than
+		// completely known. A provider that declares Complete, or that carries
+		// nodes, edges or issues, always counts toward completeness.
+		if fragment.Coverage.Status == CoverageUnavailable && len(fragment.Nodes) == 0 &&
+			len(fragment.Edges) == 0 && len(fragment.Coverage.Issues) == 0 && len(fragment.Members) == 0 {
+			sawNotApplicable = true
+			continue
+		}
 		if fragment.Coverage.Status != CoverageComplete {
 			complete = false
 		}
@@ -155,7 +167,7 @@ func MergeFragments(root string, fragments []Fragment) *Graph {
 
 	issues = uniqueSortedIssues(issues)
 	switch {
-	case len(graph.Nodes) == 0 && !complete:
+	case len(graph.Nodes) == 0 && (!complete || sawNotApplicable):
 		graph.Coverage.Status = CoverageUnavailable
 	case complete && len(issues) == 0:
 		graph.Coverage.Status = CoverageComplete
--- a/topology/graph_test.go
+++ b/topology/graph_test.go
@@ -295,3 +295,42 @@ func TestExpandReferenceRejectsAmbiguityWithoutEdges(t *testing.T) {
 		t.Fatalf("candidates = %#v, want sorted unique IDs", issue.Candidates)
 	}
 }
+
+func TestMergeFragmentsIgnoresNonApplicableProviders(t *testing.T) {
+	root := t.TempDir()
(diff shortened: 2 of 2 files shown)


JordanCoin:
CI note: **Test (ubuntu-latest, 1.24)** failed on `7ebf5a7`, with a single failure:

```
=== RUN   TestEventDetection
    watch_test.go:99: Expected positive line delta, got 0
--- FAIL: TestEventDetection (1.14s)
```

I don't believe this is this PR's, for three reasons:

1. The diff touches only `topology/graph.go` and `topology/graph_test.go` — no `watch/` code, and nothing that changes when the daemon rebuilds topology or how fsnotify events are timed. (The one adjacent concern would be cacheability, but `IsCacheable` only rejects `Unavailable` and `provider-failed`; a `partial → complete` change leaves caching behavior identical.)
2. `codemap/topology` — the package this PR actually changes — is **fully green** in the same run, including the new `TestMergeFragmentsIgnoresNonApplicableProviders`.
3. It's the known wall-clock flake family in `watch/` (cf. #135). In this very run `TestDebounce` and `TestLineDelta` — the other two members of that family — both passed, which is the signature of timing noise rather than a deterministic break. It also failed on unmodified `main` earlier today in this repo's CI, and passes 8/8 locally on `main` here.

Re-running the failed job once to confirm. If it fails again on the same test I'll treat it as real and dig in rather than assume.

---

JordanCoin:
Correction to my last comment: I can't actually re-run the job — `rerun-failed-jobs` returns `403 Resource not accessible by integration` for me on this repo. So the confirmation I promised isn't something I can perform.

Making `TestEventDetection` robust would mean changing `watch/` timing code, which is unrelated to this PR's topology change and would widen it, so I'm not doing that here. The analysis in the previous comment stands on its own evidence: `topology` is green including the new test, the diff touches no `watch/` code, and the failure is in the known wall-clock family (#135) with its two siblings passing in the same run.

If you push the re-run button and it comes back green, this is ready. If it fails again on the same test, tell me and I'll dig in properly rather than keep calling it a flake.

---

Issue #149, opened 2026-08-24
`blast-radius` truncates changed files alphabetically and presents the subset as the whole diff


JordanCoin:
**Version:** 4.4.2 (Homebrew) · macOS

## Summary

`blast-radius` caps the changed-file list at 20. Selection is by **path sort order**, not importance, and the `[diff]` panel then renders that subset with a header that reads like the complete change.

## Repro

Any branch with more than 20 changed files vs its target:

```
codemap blast-radius --text --ref <target> .
```

## Actual

```
[summary]
changed_files=20/52

[diff]
│ Changed: 20 files | +131 -112 lines vs <target> │
```

The real change was **52 files, +293 −305**. The `[diff]` header states 20 files and the reduced line counts with no indication anything was dropped.

## Which files get dropped

Selection is effectively `sorted(paths)[:20]`, so the cut lands at whatever path prefix sits at position 20 — in my case mid-alphabet. Everything after it disappeared, including:

| dropped file | churn |
|---|---|
| `.../Location/LocationManager.swift` | 62 — **the highest-churn file in the entire change** |
| `.../RootTabView.swift` | 35 |
| `.../Navigation/RouterState.swift` | 33 |
| `.../AppEntry.swift` | 33 |

So the summary omitted the largest edit, the app entry point, and the navigation root — while including small files that happened to sort earlier. For a refactor whose risk is concentrated in exactly those files, the output is actively misleading.

## Expected

Truncation is fine; silent, arbitrary truncation is not.

## Suggested fix

1. **Rank before truncating** — by churn (`added + removed`), or by hub-ness (importer count), or both. The current order carries no information.
2. **Label the truncation in the panel** — `Changed: 20 of 52 shown (ranked by churn)`. The `[summary]` block already says `20/52`; the `[diff]` header contradicts it, and that header is the part a reader quotes.
3. Consider making the cap a flag (`-max-changed-files`) that's discoverable from `blast-radius --help`.

Issue #138, opened 2026-08-15
--importers: 'No files import X' is stated for cases where the graph cannot know, and the explanation is Go-specific


JordanCoin:
Two related output bugs on the same command. Both make a *cannot-know* read as a *known-negative*, which is the answer most likely to be acted on destructively.

## 1. Same-package files report "No files import X"

codemap's graph is file-to-file via import statements. Go files in the same package do not import each other, so codemap has no edges inside a package — a structural property, not a gap in the scan. It reports it as a finding anyway:

```
$ codemap --importers watch/events.go
No files import watch/events.go.
```

`watch/events.go` is used throughout `package watch`. The honest answer is "same-package references are not modeled, I cannot tell you." What it says instead is indistinguishable from a genuinely unimported file.

Verified the asymmetry — cross-package works correctly:

```
$ codemap --importers analysis/contracts.go
⚠️  HUB FILE: analysis/contracts.go — imported by 16 files
```

So the command is reliable across packages and silently unreliable within one, with no way for the caller to tell which answer they just got.

## 2. The explanatory note is printed on non-Go files

The note under that message reads:

```
   Note: files in the same package never import each other (Go resolves
   imports at package level), so only cross-package importers appear here.
```

It is emitted unconditionally. Observed verbatim on a **TypeScript** file:

```
$ codemap --importers src/types.ts        # a NodeNext TS project
No files import src/types.ts.
   Note: files in the same package never import each other (Go resolves
   imports at package level), so only cross-package importers appear here.
```

and on a **Rust** file. Neither language has Go package semantics, and in the TypeScript case the real reason for the empty answer was #132 — a resolution bug the note actively misdirects away from.

## Suggested fix

- Scope the note to Go sources. For other languages either omit it or state the actual reason.
- Distinguish "no importers found" from "importers not modeled for this file". For a Go file whose package contains other files, say so rather than reporting zero — something like `not modeled: same-package references (N other files in package watch)`.
- Longer term this is the concrete motivating case for #133: the command should return a trust verdict alongside the count, because the count alone cannot express "I am structurally unable to answer this."

## Why it matters

`--importers` is the blast-radius command — the one an agent runs before deleting, renaming, or changing a signature. A false "no importers" is acted on. Between this and #132/#136 there are now three distinct paths to a confident zero, none of which flag themselves.

Related: #133, #132, #136, #111, #137.

Issue #111, opened 2026-08-08
coverage is reported at scan granularity but consumed at query granularity


JordanCoin:
Design issue behind #106/#107, worth separating because the fix is a contract change rather than a relabel.

## The mismatch

Coverage today is a property of the **whole scan**: "this repo contains Rust, therefore `partial`." But callers consume codemap one **query** at a time — `--importers <file>`, `get_importers`, blast radius for a specific path.

So in a Go + Rust monorepo:

```
$ codemap --importers config/config.go
⚠️  HUB FILE: config/config.go
   Imported by 22 files
Coverage: partial — Rust macro-generated, string-routed, and #[path] module edges may be unresolved
```

That caveat is true about the repo and useless about the answer. Nothing in `config/config.go`'s import graph is affected by Rust macro expansion. A caller can't tell whether the gap touches *this* result or some unrelated crate three directories away.

## Why it matters more as a primitive than as a CLI

A human skims the caveat and moves on. An agent has to make a decision: do I trust this blast-radius list enough to skip a manual check? If coverage is `partial` on essentially every query in any polyglot repo, the honest answer is always "maybe", and callers learn to ignore the field entirely. A signal that's always on is the same as no signal — and it's worse than none here, because it was added specifically to be trusted.

## Suggested direction

Make coverage answer the question the caller actually asked:

1. **Per-result coverage** — when answering for file `X`, report only the gaps that could affect `X`'s edges (its language, its resolver, the manifests in play). A Go file in a mixed repo should read `complete`.
2. **Scan-level coverage stays as a rollup** — still useful for `--deps` over a whole project, still the right shape for `DepsProject.Coverage`.
3. **Per-edge provenance where cheap** — which resolver produced each edge (`cargo-metadata`, `js-workspace`, `fuzzy`). That also makes "why does codemap think A imports B?" answerable, which is currently guesswork.

## Relationship to the open issues

- #106 (ast-grep mislabeled `mixed` with a Rust caveat) is the sharpest instance: an unrelated source gets tagged degraded because Rust exists somewhere in the tree. Fixing #106 narrows the blast radius of the mislabel but does not fix the granularity.
- #107 (duplicate note) is downstream of #106.

Both are worth fixing on their own; this issue is the reason they keep recurring.

## Reclassification note

[...]

PR #167, opened 2026-09-02
feat(context): Resolve bounded lexical file mentions

reneleonhardt:
## What does this PR do?

- Resolves exact paths, unique extension basenames, unique stems, and configured subsystem paths.
- Normalizes repository-relative paths across slash styles and rejects traversal, absolute, volume, or ambiguous case-folded matches.
- Feeds resolved files into the existing evidence-backed intent risk model.

This makes common agentic coding prompts such as `refactor final_build` useful without fuzzy dependency false positives.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other

## Checklist

- [x] Tested locally with Go 1.27.0: `go test ./...`
- [x] Verified traversal, ambiguity, case, duplicate subsystem, and top-k boundaries
- [x] Verified with `go vet ./...`

## Additional notes

Routing is inventory-only and performs no semantic search or unbounded expansion.

### DIFF
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 	"time"
@@ -147,8 +148,9 @@ func buildContextEnvelopeWithDeps(parent context.Context, root, prompt string, c
 	// Intent classification (if prompt provided)
 	if prompt != "" {
 		topK := projCfg.RoutingTopKOrDefault()
-		files := resolveContextFiles(prompt, inputs.files, projCfg, topK)
-		intent := classifyIntent(prompt, files, inputs.info, projCfg)
+		resolution := resolveContextFileResolutionWithCase(prompt, inputs.files, projCfg, topK, runtime.GOOS == "windows")
+		files := resolution.files
+		intent := classifyIntentWithResolution(prompt, resolution, inputs.info, projCfg)
 		envelope.Intent = &intent
 
 		// Match skills against intent
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -147,8 +147,7 @@ func buildContextEnvelopeWithDeps(parent context.Context, root, prompt string, c
 	// Intent classification (if prompt provided)
 	if prompt != "" {
 		topK := projCfg.RoutingTopKOrDefault()
-		mentions := extractMentionedFiles(strings.ReplaceAll(prompt, `\`, "/"), topK)
-		files := resolveExactConfiguredFiles(mentions, inputs.fileSet)
+		files := resolveContextFiles(prompt, inputs.files, projCfg, topK)
 		intent := classifyIntent(prompt, files, inputs.info, projCfg)
 		envelope.Intent = &intent
 
--- a/cmd/context_evidence.go
+++ b/cmd/context_evidence.go
@@ -149,7 +149,7 @@ func sortedGraphHubs(importers map[string][]string) []string {
 }
 
 func normalizeContextPath(file string) string {
-	return normalizeContextPathWithVolumeGuard(file, true)
+	return normalizeContextPathWithVolumeGuard(strings.TrimSpace(file), true)
 }
 
 func normalizeContextInventoryPath(file string) string {
@@ -157,7 +157,7 @@ func normalizeContextInventoryPath(file string) string {
 }
 
 func normalizeContextPathWithVolumeGuard(file string, rejectVolume bool) string {
-	file = strings.TrimSpace(strings.ReplaceAll(file, `\`, "/"))
+	file = strings.ReplaceAll(file, `\`, "/")
 	volumePath := isContextVolumePath(file)
 	file = strings.TrimPrefix(pathpkg.Clean(file), "./")
 	if file == "." || file == "" || strings.HasPrefix(file, "../") || file == ".." || pathpkg.IsAbs(file) || (rejectVolume && volumePath) {
--- a/cmd/context_evidence.go
+++ b/cmd/context_evidence.go
@@ -44,7 +44,6 @@ type contextEnvelopeDeps struct {
 type contextRequestInputs struct {
 	state      *watch.State
 	files      []scanner.FileInfo
-	fileSet    map[string]string
 	info       *hubInfo
 	evidence   GraphEvidence
 	fileScanOK bool
@@ -67,7 +66,6 @@ func defaultContextEnvelopeDeps() contextEnvelopeDeps {
 func loadContextRequestInputs(ctx context.Context, root string, deps contextEnvelopeDeps) contextRequestInputs {
 	inputs := contextRequestInputs{
 		state:    deps.readState(root),
-		fileSet:  make(map[string]string),
 		evidence: unavailableGraphEvidence(graphEvidenceScanFailed),
 	}
 
@@ -78,12 +76,6 @@ func loadContextRequestInputs(ctx context.Context, root string, deps contextEnve
 	}
 	inputs.files = files
 	inputs.fileScanOK = true
-	for _, file := range files {
-		normalized := normalizeContextPath(file.Path)
-		if normalized != "" {
-			inputs.fileSet[normalized] = file.Path
-		}
-	}
 
 	if len(files) == 0 {
 		inputs.info = &hubInfo{Importers: map[string][]string{}, Imports: map[string][]string{}}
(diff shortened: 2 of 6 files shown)


JordanCoin:
Reviewed this closely and built binaries from both `f9ed58d` (main) and `c030bb9` to compare. The exact-path and basename passes look right to me, and the `len(matches) == 1` ambiguity guards are genuinely load-bearing — I mutated each to `>= 1` and the corresponding test failed every time. Nice.

The problem is what passes 3 and 4 do to `TaskIntent`.

## Blocking: passes 3 and 4 manufacture a risk verdict from prompts that mention no file

`TaskIntent.Files` is documented in `cmd/intent.go:15` as `// mentioned files`, and it is the sole gate on `analyzeRisk` — `classifyIntent` only computes a risk level when `len(files) > 0`. So anything that puts a file in that list converts an honest `risk: unknown` into a confident one.

**Pass 4 (subsystem paths).** With `{id: scanner, keywords: [parsing], paths: [scanner]}` configured and the prompt `"why is parsing slow?"`:

```
main : files=[]                                             scope=unknown  risk=unknown
#167 : files=[scanner/alpha.go beta.go delta.go epsilon.go] scope=package  risk=low
```

Four files chosen by alphabetical position and truncated at topK — `gamma.go` and `zeta.go` lost a coin flip, not a relevance contest. This is the same shape as #149, where `blast-radius` truncates alphabetically and presents the subset as the whole.

**Pass 3 (stem matching) is worse, because it needs no config at all.** `contextRoutingTokens` tokenizes the entire prompt with no stopword filter, so any ordinary English word that happens to be a unique file stem resolves. Run against codemap's own repo:

```
"I need a doctor's note before I can travel"
  main : files=[]                    scope=unknown      risk=unknown
  #167 : files=['cmd/doctor.go']     scope=single-file  risk=low

"please install a new espresso machine in the kitchen"
  main : files=[]                    scope=unknown      risk=unknown
  #167 : files=['plugins/install.go'] scope=single-file risk=low

"let's root out the flaky tests in CI"
  main : files=[]                    scope=unknown      risk=unknown
  #167 : files=['cmd/root.go']       scope=single-file  risk=low

"the build was slow again today"
  main : files=[]                    scope=unknown      risk=unknown
  #167 : files=['handoff/build.go']  scope=single-file  risk=low
```

`root`, `setup`, `serve`, `drift`, `doctor`, `install`, `build` and `release` all have unique stems in this repo. An agent reading `risk: low` on the espresso prompt has been told something false, and nothing in `TaskIntent` distinguishes "the user named this file" from "the resolver guessed it from a common word" — same field, same shape, no provenance marker. That's the distinction #138 and #111 are both circling.

The fix I'd suggest, smallest first: drop pass 3 (require an extension, keeping the basename pass), and thread a `fuzzy`/`confidence` marker through to `TaskIntent` so `computeScope` and `analyzeRisk` return `unknown` when guessed files are the only evidence. A guessed file is fine as a *suggestion*; it is not fine as the basis of a risk verdict.

## Blocking: the two cross-cutting guarantees in the description have no test coverage

Both verified by mutation, `-count=1` so nothing is cached:

- **Volume-path rejection.** I replaced `|| isContextVolumePath(file)` with `|| (false && isContextVolumePath(file))` — neutering the new guard entirely — and the whole `TestContextLexicalRouting` suite still passes, *including* the subtest named `absolute and volume paths stay unresolved`. That test's assertions hold because `path.Clean` collapses `//server/...` to a single leading `/` (so `pathpkg.IsAbs` short-circuits first) and because `contextRoutingTokenPattern` excludes `:`, so `C:\tmp\target.go` tokenizes to `["C", "/tmp/target.go"]` and the path half is `IsAbs`-rejected on its own. The guard's `"//"` branch is unreachable dead code, and its drive-letter branch is unreachable from prompt tokens — it only does real work on `subsystem.Paths` and on scanner paths in `newContextFileIndex`.
- **Pass ordering.** Swapping pass 3 above pass 1 in the function body leaves all 13 subtests green, including `exact path wins before stem` — its fixture has the stem and the exact path resolving to the *same* file, so it cannot detect a priority regression. The shipped code does implement the priority correctly (I checked with topK=1 and reversed token order), but nothing protects it.

Worth adding: a fixture where pass 1 and pass 3 resolve to *different* files, and a direct `normalizeContextPath` call that actually reaches the drive-letter branch.

## Smaller things

- `contextSubsystemMatches` is a near-verbatim copy of `matchSubsystemRoutes` (`cmd/hooks.go:1275`) — same scoring, same tie-break. Two copies will drift. It's also called with `len(cfg.Routing.Subsystems)` instead of `RoutingTopKOrDefault()`, so file routing considers more subsystems than `intent.Subsystems` reports, and a file can appear with no visible reason.
- `contextRequestInputs.fileSet` (`cmd/context_evidence.go:47`) is still built on every call but no longer read by any non-test code.
- Because `newContextFileIndex` runs real scanner paths through the volume guard, a legitimate POSIX file like `x:helper.go` is dropped from `exact`, `basenames`, `stems` and `paths` — unfindable even by a verbatim mention. Narrow, and fail-closed rather than wrong, but easy to avoid by applying the guard only to prompt tokens and config-authored paths.
- Pass 4 iterates every indexed path per prefix per matched subsystem. Measured ~89ms for 40 subsystems × 5 prefixes × 20k files, on top of ~21ms index construction. `context` is on the editor-hook hot path; the topK mismatch above compounds it.

The exact and basename passes are a real improvement and I'd take those on their own. It's the confidence that passes 3 and 4 manufacture that I can't merge — `risk: low` on a prompt about an espresso machine is the failure mode this project exists to avoid. Happy to be argued out of any of it if I've misread the intent.

Baseline note: `go test ./cmd/...` on this branch is clean apart from `TestRunSetupCreatesConfigAndHooks`, which fails for anyone running as root.

---

reneleonhardt:
### Complied

- Removed extensionless stem inference.
- Added file-confidence provenance.
- Prevented inferred candidates from driving scope or risk.
- Added the requested volume-path and ordering coverage.

### Other changes

- Fixed mixed explicit/inferred evidence handling found during our adversarial loop.
- Made duplicate subsystem ordering deterministic.
- Removed dead `fileSet` state.
- Fixed significant-whitespace inventory paths and added regression coverage.

JordanCoin:
Re-verified everything at `8cd19c2` with fresh binaries. All of it holds up.

**The fabrication is gone.** Same prompts as before, against codemap's own repo:

```
"please install a new espresso machine in the kitchen"  files=[]  scope=unknown  risk=unknown
"I need a doctor's note before I can travel"            files=[]  scope=unknown  risk=unknown
"let's root out the flaky tests in CI"                  files=[]  scope=unknown  risk=unknown
"the build was slow again today"                        files=[]  scope=unknown  risk=unknown
"inspect cmd/context.go for the bug"   files=['cmd/context.go']  scope=single-file  risk=low
```

**The `file_confidence` design is better than what I suggested.** Inferred candidates are still surfaced — which is the useful half — but honestly labelled and firewalled from the verdict. The subsystem case now reads:

```json
"files": ["scanner/alpha.go", "scanner/beta.go", "scanner/delta.go", "scanner/epsilon.go"],
"file_confidence": "inferred",
"scope": "unknown",
"risk": "unknown",
"suggestions": [{"reason": "risk is unknown without an explicit file mention"}]
```

and the mixed case computes scope and risk from the explicit file alone while still listing the inferred ones as `"file_confidence": "mixed"`. That's the right shape — I'd only floated dropping the candidates entirely, and keeping them with provenance is more useful.

**The new tests are load-bearing.** I re-ran the mutations that previously passed:

- Neutering the volume guard (`rejectVolume && volumePath` → `false && ...`) now **fails** `absolute_and_volume_paths_stay_unresolved`. Hoisting `isContextVolumePath` above `pathpkg.Clean` also un-deadens the `"//"` branch, which was the underlying reason the old test couldn't see it.
- Mislabelling exact matches as inferred **fails** `inferred_candidates_do_not_affect_explicit_scope_or_risk` and `TestContextExactPathResolution/separator_normalized`.
- Labelling subsystem candidates as explicit **fails** both `inferred_*_do_not_set_scope_or_risk` tests.

Also confirmed fixed: the `RoutingTopKOrDefault()` mismatch, the dead `fileSet`, and the real-scanner-path rejection (`normalizeContextInventoryPath` keeps files like `x:helper.go` and whitespace-significant names in the index).

`go vet ./...` clean; `go test ./cmd/... ./config/...` at baseline (only `TestRunSetupCreatesConfigAndHooks`, which fails for anyone running as root). CI 12/12.

Two things I'm deliberately not holding this for, noting them for later: `contextSubsystemMatches` still duplicates `matchSubsystemRoutes` scoring, and the subsystem pass is still O(subsystems × prefixes × files). Neither affects correctness.

Merging. Thanks for the quick turnaround on this one — the provenance field is a genuine improvement to the contract, not just a patch over my complaint.

---

Issue #148, opened 2026-08-24
`--deps` / `--importers` / `blast-radius` report `coverage: complete` on languages with no file-level imports


JordanCoin:
**Version:** 4.4.2 (Homebrew) · macOS · `ast-grep` installed

## Summary

In languages where files inside a module do not import each other by path — Swift is the clearest case, but the same applies to C#, Java within a package, and Kotlin — the dependency graph comes back essentially empty, yet the JSON still declares the analysis authoritative and complete.

## Repro

Any single-module Swift app (~290 source files in my test):

```
codemap --deps --json .
```

## Actual

```json
"coverage": {"status": "complete", "sources": [{"name": "ast-grep", "status": "authoritative"}], "issues": []}
```

Two edges found across the whole project. `blast-radius` on a 50-file change in the same project reports:

```
files_with_dependents=0
impacted_outside_diff=0/0
dependency_context_outside_diff=0/0
```

## Expected

Either real edges, or an honest `degraded` status.

## Why this is the worst of the bugs

Parsing is fine — 1068 functions were extracted correctly. The *edge model* is what doesn't apply: every import in that project resolves to a framework (`SwiftUI`, `Foundation`, …), never to another file in the project. So the graph is structurally empty, not wrong-by-accident.

The damage comes from pairing that with `status: complete`. A consumer — especially an automated one feeding an LLM — reads "complete" plus "0 dependents" and concludes the change is isolated. In my test the changed set included a type with **40** referencing files and another with **33**. Reported impact: zero, with full confidence.

A tool that says "I found nothing" is useful. A tool that says "I checked everything and there is nothing" when it cannot check is worse than not running.

## Suggested fix

1. When a language's import graph yields ~0 internal edges across a large file count, set `coverage.status = "degraded"` with an issue explaining that the language needs symbol-reference resolution rather than import resolution.
2. Longer term, resolve edges for these languages by **type name → declaring file**. `ast-grep` already extracts the declarations; the missing step is matching references to them. A plain ranking of `type name → count of other files referencing it` took 1.8s on the same repo and produced exactly the signal `blast-radius` reported as zero.

## Related false positives

The two edges it does find look like filename-against-directory-name matches:

```
VendorSDK  ───▶ src/Integrations/VendorSDK
OtherSDK   ───▶ src/Integrations/OtherSDK
```

PR #119, opened 2026-08-08
fix(context): Report bounded graph evidence honestly

reneleonhardt:
## What does this PR do?

- Builds one bounded request-local dependency graph for CLI and HTTP context.
- Reports unavailable evidence and unknown risk instead of false zero hubs or low risk.
- Keeps prompt hooks scan-free and propagates HTTP cancellation.

This prevents agentic coding workflows from treating missing graph evidence as a safe change.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [x] Documentation
- [ ] Other

## Checklist

- [x] Tested locally with `GOENV_VERSION=1.26.5 go test ./cmd ./scanner ./watch -count=1`
- [x] Verified with full vet and race tests
- [x] Updated context and hook documentation

## Additional notes

The JSON envelope is version 2: hub counts are nullable and graph availability is explicit.

Developed with carefully directed, manually reviewed AI assistance.

PR #168, opened 2026-09-02
fix(watch): Validate cached graph provenance

reneleonhardt:
## What does this PR do?

- Persists graph lifecycle, project, filter, builder, and exact inventory provenance.
- Coalesces duplicate writes, persists configured membership changes, and publishes rebuilt graphs atomically.
- Reuses cached evidence only when fresh context inventory proves an exact match.

This prevents agentic coding workflows from relying on stale dependency data after edits, filter changes, or membership changes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New language support
- [ ] Documentation
- [ ] Other

## Checklist

- [x] Tested locally with Go 1.27.0: `go test ./...`
- [x] Race-tested watch publication and event paths: `go test -race ./...`
- [x] Verified with `go vet ./...`

## Additional notes

Hooks remain scan-free; context performs the exact current-inventory proof before cache reuse.

### DIFF
--- a/cmd/context_evidence.go
+++ b/cmd/context_evidence.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"codemap/analysis"
 	"codemap/config"
 	"codemap/limits"
 	"codemap/scanner"
@@ -105,7 +106,13 @@ func loadContextRequestInputs(ctx context.Context, root string, deps contextEnve
 		inputs.evidence = unavailableGraphEvidence(graphErrorReason(ctx, err))
 		return inputs
 	}
-	if graph == nil || (len(graph.Imports) == 0 && len(graph.Importers) == 0) {
+	if graph == nil {
+		inputs.evidence = unavailableGraphEvidence(graphEvidenceScanIncomplete)
+		return inputs
+	}
+	// Complete edge-free graphs are valid; unavailable provenance is not.
+	if graph.Coverage.Status == analysis.CoverageUnavailable ||
+		(len(graph.Coverage.Sources) > 0 && scanner.CoverageFromSources(graph.Coverage.Sources).Status == analysis.CoverageUnavailable) {
 		inputs.evidence = unavailableGraphEvidence(graphEvidenceScanIncomplete)
 		return inputs
 	}
--- a/cmd/context_evidence.go
+++ b/cmd/context_evidence.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"time"
 
-	"codemap/analysis"
+	"codemap/config"
 	"codemap/limits"
 	"codemap/scanner"
 	"codemap/watch"
@@ -18,6 +18,7 @@ const (
 	graphEvidenceAvailable       = "available"
 	graphEvidenceUnavailable     = "unavailable"
 	graphEvidenceFreshScan       = "fresh_scan"
+	graphEvidenceWatchCache      = "watch_cache"
 	graphEvidenceLargeRepository = "large_repository"
 	graphEvidenceCancelled       = "cancelled"
 	graphEvidenceDeadline        = "deadline"
@@ -90,6 +91,10 @@ func loadContextRequestInputs(ctx context.Context, root string, deps contextEnve
 		inputs.evidence = GraphEvidence{Status: graphEvidenceAvailable, Source: graphEvidenceFreshScan}
 		return inputs
 	}
+	if graph, _ := watch.ValidateCachedGraphForInventory(inputs.state, root, config.Load(root), contextFilePaths(files)); graph != nil {
+		populateContextGraphInputs(&inputs, graph, graphEvidenceWatchCache)
+		return inputs
+	}
 	if len(files) > limits.LargeRepoFileCount {
 		inputs.evidence = unavailableGraphEvidence(graphEvidenceLargeRepository)
 		return inputs
@@ -100,20 +105,24 @@ func loadContextRequestInputs(ctx context.Context, root string, deps contextEnve
 		inputs.evidence = unavailableGraphEvidence(graphErrorReason(ctx, err))
 		return inputs
 	}
-	if graph == nil {
-		inputs.evidence = unavailableGraphEvidence(graphEvidenceScanIncomplete)
-		return inputs
-	}
-	// An edge-free graph is still authoritative when its coverage is complete
-	// (the zero coverage status is the production graph's complete default).
-	// Explicitly unavailable provenance remains fail-closed, even if stale or
-	// partial edge maps happen to contain entries.
-	if graph.Coverage.Status == analysis.CoverageUnavailable ||
-		(len(graph.Coverage.Sources) > 0 && scanner.CoverageFromSources(graph.Coverage.Sources).Status == analysis.CoverageUnavailable) {
+	if graph == nil || (len(graph.Imports) == 0 && len(graph.Importers) == 0) {
 		inputs.evidence = unavailableGraphEvidence(graphEvidenceScanIncomplete)
 		return inputs
 	}
 
+	populateContextGraphInputs(&inputs, graph, graphEvidenceFreshScan)
+	return inputs
+}
+
+func contextFilePaths(files []scanner.FileInfo) []string {
+	paths := make([]string, 0, len(files))
+	for _, file := range files {
+		paths = append(paths, file.Path)
+	}
+	return paths
+}
+
+func populateContextGraphInputs(inputs *contextRequestInputs, graph *scanner.FileGraph, source string) {
 	hubs := sortedGraphHubs(graph.Importers)
 	inputs.info = &hubInfo{
 		Hubs:      hubs,
@@ -121,8 +130,7 @@ func loadContextRequestInputs(ctx context.Context, root string, deps contextEnve
 		Imports:   graph.Imports,
 		Coverage:  graph.Coverage,
 	}
-	inputs.evidence = GraphEvidence{Status: graphEvidenceAvailable, Source: graphEvidenceFreshScan}
-	return inputs
+	inputs.evidence = GraphEvidence{Status: graphEvidenceAvailable, Source: source}
 }
 
 func unavailableGraphEvidence(reason string) GraphEvidence {
--- a/cmd/context_evidence_test.go
+++ b/cmd/context_evidence_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"codemap/analysis"
 	"codemap/config"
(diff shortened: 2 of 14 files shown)


JordanCoin:
Spent a while on this one — it's a good idea and the provenance record (`GraphState` with root, filter, builder and inventory fingerprints) is the right shape. But I reproduced three problems I can't merge past, each against a binary built from `f9ed58d` and one from `85174ad`.

## Blocking 1: the first save to any source file permanently kills the daemon's graph

`markGraphLifecycleLocked` sets `FileGraph = nil`, `HasDeps = false` and clears `DepCtx`, and `handleEvent` now calls it on every create/write/delete of a configured file. But `computeDeps()` is only reachable from `computeInitialGraphs` at startup and from `refreshConfiguredFiles`, which the event loop invokes **only for control events** (`.gitignore` / `.codemap/config.json`, via `filterControlEvent` → `armControlTimer` → `refreshConfigured`). Nothing schedules a rebuild after an ordinary file event.

Live daemon on a three-file Go repo with real cross-package imports:

```
t=4s   startup:                     importers=2  imports=2  status=available
t=7s   after ONE comment-only edit: importers=0  imports=0  status=stale
t=22s  ...                          importers=0  imports=0  status=stale
t=37s  ...                          importers=0  imports=0  status=stale
--- touch .codemap/config.json (a control event) ---
t=42s                               importers=2  imports=2  status=available
```

Main holds `importers=2` throughout. So the daemon goes dark on your first keystroke-save and stays dark for the rest of its life, and the only thing that revives it is editing the config. That's the opposite of the PR's goal — the watch daemon exists precisely to keep this warm *while* you edit. It doesn't serve stale data; it serves nothing, forever.

`TestConfiguredEventInvalidatesGraph` asserts the "goes stale" half and never asserts recovery, so it encodes this as intended.

## Blocking 2: the fail-closed coverage guard from #119 is inverted in both directions

`loadContextRequestInputs` drops

```go
if graph.Coverage.Status == analysis.CoverageUnavailable ||
    (len(graph.Coverage.Sources) > 0 && scanner.CoverageFromSources(graph.Coverage.Sources).Status == analysis.CoverageUnavailable) {
```

in favour of `if graph == nil || (len(graph.Imports) == 0 && len(graph.Importers) == 0)`. `git log -S` puts that block at `71f60e4` — your own #119, *"Report bounded graph evidence honestly"*, whose second commit was literally *"distinguish edge-free complete graphs"*.

Injecting graphs through the `buildFileGraph` seam and reading back `inputs.evidence`:

| graph | main (f9ed58d) | this PR |
|---|---|---|
| coverage=**unavailable**, has edges | `unavailable` / `scan_incomplete`, `info=nil` | **`available` / `fresh_scan`, `info` populated** |
| coverage=**complete**, edge-free | `available` / `fresh_scan`, `info` populated | **`unavailable` / `scan_incomplete`, `info=nil`** |

Row 1 is the cardinal-rule violation: a graph that marked its own coverage unavailable now reaches `analyzeRisk`, which indexes `info.Importers[file]` and emits confident `risk` plus `review-hub` / `check-deps` advice from data the graph itself disclaimed. That's exactly what the old comment ("even if stale or partial edge maps happen to contain entries") was guarding. Row 2 is the milder half — every leaf project with no cross-file imports now reports a false `scan_incomplete`. Related to #111 and #148.

`ValidateCachedGraph` has the same gap: it returns a `*scanner.FileGraph` without ever inspecting `Coverage.Status`.

## Blocking 3: `hook prompt-submit` now emits a confident risk level from a content-unvalidated cache

`getValidatedHubInfo` calls `ValidateCachedGraph`, **not** `ValidateCachedGraphForInventory` — so the exact configured-path-set check the PR description advertises is skipped at this call site. And `ConfiguredInventoryFingerprint` is over paths only: no content hash, no mtime. An edit that changes a file's imports without adding or removing a file is invisible to it.

Populated a state file, stopped the daemon, then removed an import from `pkgb/b.go` (path set unchanged; `pkga/a.go` truly drops from 2 importers to 1):

```
main :  "risk":"unknown"  suggestions: [check-deps . — "risk is unknown without a
                                        configured file and available dependency graph"]
#168 :  "risk":"low"      suggestions: [run-tests . — "run full test suite after refactoring"]
```

To be fair about what is and isn't new here: **both** builds print `📍 pkga/a.go (imported by 2 files)`, which is wrong on both — main's `getHubInfoWithFallback` reads `state.Importers` with no validation at all, so that staleness is pre-existing (#140's family), not something this PR introduced. What *is* new is that main passed `nil` into `classifyIntent`, so the hook could never derive a risk verdict from cached data and correctly said `unknown`. This PR passes `info` in, and the honest "risk is unknown" note becomes a confident `risk: low`. Given the PR's headline is "prevents relying on stale dependency data after edits", turning an honest `unknown` into a confident `low` from a cache that cannot detect edits is the wrong direction.

## Things I checked that are fine

- The `findRelatedHot` rewrite is behavior-preserving — `ConnectedFiles(path)` is exactly `Imports[path] ∪ Importers[path]` minus self, dedup is preserved via delete-on-match, and the new newest-first walk is *more* deterministic than the old map iteration.
- No lock-ordering problem: `markGraphLifecycle` (which takes the lock) is never called under it. `go build -race ./...` and `go test -race ./watch/... ./cmd/...` are clean apart from the three baseline root-permission failures.

## Smaller

- `computeDepsWithBeforePublish` discards a freshly built graph when a racing event bumps `graphGeneration`, and marks stale without scheduling a retry — same shape as blocking 1.
- The `os.IsNotExist` race branch in `handleEvent` now calls `writeState()` unconditionally, even when `!wasConfigured`, bypassing the publisher's 100ms coalescing.
- `resolveValidatedHookFiles` rejects symlinks via `os.Lstat` + `IsRegular()`, stricter than `ScanConfiguredFiles` — deliberate? (The gitignore warm-up loop is fine; `normalizeContextPath` already rejects `../`, absolute and volume paths, so it can't escape root.)
- `validateCachedGraph` (lowercase) is used only by tests. `newGraphState` and the lowercase `graphLifecycle*` consts *are* used in production, so those are fine.

Roughly what would unblock it for me: rebuild after an ordinary file event instead of only on control events (or make `stale` trigger a scan on next read rather than serving silence); restore the coverage-status check alongside the empty-edge check in both `loadContextRequestInputs` and `ValidateCachedGraph`; and either point `getValidatedHubInfo` at `ValidateCachedGraphForInventory` or keep passing `nil` into `classifyIntent` until the cache can actually detect content changes.

Happy to be wrong on any of this — the repro repos are trivial to rebuild if you want to check.

---

reneleonhardt:
### Complied

- Configured file events now rebuild and republish dependency state.
- Unavailable coverage is rejected; complete edge-free graphs remain valid.
- Prompt-submit risk stays unknown when cache freshness cannot be proven.
- Delayed duplicate writes remain suppressed without skipping lifecycle events.

### Not fully complied

- Cache freshness remains path-based, so cached risk cannot be safely restored until content fingerprints exist.
- Generation-change races remain fail-closed as stale without an independent retry worker.

### Other changes

- Added regression coverage for recovery, coverage states, delayed writes, lifecycle events, and conservative hook behavior.
- Removed publication writes for unconfigured stat-race events.

### Content fingerprints

Feasible, but requires hashing configured files during graph publication and validating those hashes before hook-submit risk analysis. Size/mtime can avoid most rehashing; changed or ambiguous metadata must be hashed.

### Independent retry worker

Feasible, but it must build immutable results off-thread and send them back to the event loop. The event loop must remain the sole owner of `Graph`, `WorkingSet`, and the publisher, applying results only when the generation still matches.

JordanCoin:
Re-ran all three repros at `fbcdcad`. All three are fixed, and I appreciate you separating "not fully complied" out rather than papering over it.

**Blocking 1 — daemon recovery.** Same live daemon, same scratch repo:

```
t=4s   startup:                      importers=2  imports=2  status=available
t=7s   after one source edit:        importers=2  imports=2  status=available
t=12s:                               importers=2  imports=2  status=available
t=22s:                               importers=2  imports=2  status=available
```

I also checked the rebuild produces *correct* data rather than just staying nominally `available`, which turned out to be more interesting than I expected:

```
                                     truth      this PR              main
baseline                             2          [main.go, pkgb/b.go] [main.go, pkgb/b.go]
drop pkgb's import of pkga           1          [main.go]            [main.go, pkgb/b.go]
add pkgb/extra.go importing pkga     2          [.., pkgb/extra.go]  [main.go, pkgb/b.go]
add new pkgc/ dir + file             3          [.., pkgc/c.go]      [main.go, pkgb/b.go]
```

Main never refreshes at all — it serves the startup snapshot for the daemon's whole life. So this doesn't just restore the old behavior, it fixes a staleness bug that was already there. (One note for anyone reproducing: creating a directory and a file in it in the same instant races the watcher registration, so pause between the two. That's fsnotify, not this PR — I checked.)

**Blocking 2 — coverage guard.** Injecting through the `buildFileGraph` seam, both rows now match main:

```
coverage=unavailable, has edges -> unavailable / scan_incomplete, info=nil   [fails closed]
coverage=complete, edge-free    -> available   / fresh_scan,      info!=nil  [stays available]
```

`ValidateCachedGraph` picks up the same check, including the `CoverageFromSources` fallback when the status is unset.

**Blocking 3 — prompt-submit risk.** Daemon stopped, import removed in place (cache says 2 importers, truth is 1):

```
main :  "risk":"unknown"  [check-deps . — "risk is unknown without a configured file and available dependency graph"]
#168 :  "risk":"unknown"  [check-deps . — "risk is unknown without a configured file and available dependency graph"]
```

Identical to main. The comment on the `nil` explains exactly why, which is the right thing to leave behind for whoever adds content fingerprints later.

**On your two "not fully complied" items** — both are the right call. Path-only freshness genuinely cannot support a risk verdict, and saying so in the code beats pretending otherwise; that's the whole house rule. Your sketch of the retry worker (build off-thread, apply only when the generation still matches, event loop stays sole owner of `Graph`/`WorkingSet`/publisher) is the correct shape. Worth noting that `refreshDependencies` currently runs `computeDeps()` synchronously on the event loop, so on a mid-size repo a save blocks event processing for the length of a full `BuildFileGraph` — that's the concrete motivation for the worker when you get to it. Not blocking; `shouldComputeDependencyGraph` keeps large repos out of it.

Verification: `go vet ./...` clean. `go test -race ./watch/...` 5/5 clean. Full `go test ./...` at baseline (only the three root-permission failures). CI 12/12. I also merged this against #167 locally first — clean merge, combined tree builds, vets, and tests at baseline, with #167's `file_confidence` behavior intact.

Merging. Good work on this one — the diagnosis in your response matched what I measured on every point.

---

PR #169, opened 2026-09-03
perf(context): Optimize subsystem prefix routing

reneleonhardt:
# perf(context): Optimize subsystem prefix routing

## What does this PR do?

- Reuses `matchSubsystemRoutes` for context routing, keeping scoring, ordering, and top-K selection in one place.
- Builds a sorted normalized inventory index and uses binary-searched prefix ranges so subsystem routing visits only matching files.
- Replaces repeated subsystem-prefix scans and per-directory prefix materialization with index construction plus matched candidates.
- Skips duplicate configured prefixes while preserving prior path order, including bounded top-K results on case-insensitive platforms.
- Keeps exact-path and basename resolution ahead of bounded subsystem suggestions.

This reduces subsystem routing work without changing its matching or confidence semantics.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] New language support
- [ ] Documentation
- [x] Other (performance and maintainability)

## Checklist

- [x] Tested locally with Go 1.27.1: `go test ./...`
- [x] Verified focused routing tests, including ordering, top-K, case handling, duplicate prefixes, and prefix boundaries.
- [x] Verified focused routing tests under `-race`.
- [x] Verified with `go vet ./...`.
- [x] Updated the relevant unit coverage for shared route scoring and prefix indexing.

## Additional notes

Routing remains inventory-only, bounded, deterministic, and free of semantic search or unbounded expansion.

On a stress fixture with 40 matched subsystems, five prefixes each, and 20,000 files, the sorted-range resolver measured 4.3 ms versus 14.1 ms for the previous prefix-map implementation (~3.3× faster), with 61% fewer allocated bytes and 80% fewer allocations.

### DIFF
--- a/cmd/context_routing.go
+++ b/cmd/context_routing.go
@@ -14,9 +14,15 @@ var contextRoutingTokenPattern = regexp.MustCompile(`[A-Za-z0-9_@./\\-]*[A-Za-z0
 
 type contextFileIndex struct {
 	caseInsensitive bool
-	paths           []string
 	exact           map[string][]string
 	basenames       map[string][]string
+	sortedPaths     []contextIndexedPath
+}
+
+type contextIndexedPath struct {
+	path  string
+	key   string
+	order int
 }
 
 type contextFileResolution struct {
@@ -80,6 +86,7 @@ func resolveContextFileResolutionWithCase(prompt string, files []scanner.FileInf
 	}
 
 	// Configured subsystem routes are bounded last-resort file candidates.
+	seenPrefixes := make(map[string]struct{})
 	for _, subsystemIndex := range contextSubsystemMatches(prompt, cfg, cfg.RoutingTopKOrDefault()) {
 		subsystem := cfg.Routing.Subsystems[subsystemIndex]
 		for _, prefix := range subsystem.Paths {
@@ -88,11 +95,14 @@ func resolveContextFileResolutionWithCase(prompt string, files []scanner.FileInf
 				continue
 			}
 			prefixKey := index.key(prefix)
-			for _, path := range index.paths {
-				pathKey := index.key(path)
-				if (pathKey == prefixKey || strings.HasPrefix(pathKey, prefixKey+"/")) && add(path, true) {
-					return resolution
-				}
+			if _, duplicate := seenPrefixes[prefixKey]; duplicate {
+				continue
+			}
+			seenPrefixes[prefixKey] = struct{}{}
+			if index.forPrefix(prefixKey, func(path string) bool {
+				return add(path, true)
+			}) {
+				return resolution
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


JordanCoin:
Reviewed at `6f6e9ce`. The duplication fix and the sorted-range resolver both hold up, and I went after the ordering semantics hard because binary-searched prefix ranges are exactly where this kind of optimization usually changes behavior quietly.

**Behavior is identical to main.** Built binaries from `b713d5f` and `6f6e9ce` and ran them against a fixture designed to break the prefix boundary — overlapping subsystems, a duplicate prefix listed twice, `scanner/deep/` nested under `scanner/`, a `scannerx/` sibling, and a file named `scanner.go` sitting next to the `scanner/` directory:

```
"why is parsing slow?"              SAME
"parsing and scanner/alpha.go"      SAME
"parsing scanner.go"                SAME
"look at scannerx"                  SAME
```

`scanner.go` is the interesting one — `.` (0x2E) sorts below `/` (0x2F), so a naive range scan from `prefixKey` would sweep it up. The separate exact-key range plus `sort.Search` on `prefixKey + "/"` gets it right, and `scannerx/x.go` is correctly excluded too.

**The case-insensitive path preserves emission order.** This is the subtle part: `sortedPaths` is sorted by folded key for binary search, but the old code emitted in raw-path `sort.Strings` order. Stashing `order` from a pre-sort and re-sorting matches by it does reproduce that. Verified directly:

```
files:  Scanner/Zeta.go scanner/alpha.go SCANNER/Beta.go
        scanner/deep/Nu.go Scanner/Deep/delta.go scanner/mu.go
prefix "scanner", caseInsensitive=true
got:    [SCANNER/Beta.go Scanner/Deep/delta.go Scanner/Zeta.go
         scanner/alpha.go scanner/deep/Nu.go scanner/mu.go]
        == sort.Strings(raw)  ✓
```

**`contextSubsystemMatches` now delegates to `matchSubsystemRoutes`.** That closes the duplicated-scoring finding from #167 properly — one scorer, one tie-break, and the `Index` field is a cleaner way to bridge it than what I'd have written.

`go vet ./...` clean, `go test ./...` at baseline (only the three root-permission failures).

## On the red CI

Both failing legs are in packages this diff cannot reach, and `codemap/cmd` — the only package it touches — passed on **both** of them:

- `Test (ubuntu-latest, 1.26)`: `mcp/TestRustGraphContextHandlersDisclosePartialCoverage` — `Failed to build file graph: context deadline exceeded`. A timeout building a Rust graph in `codemap/mcp`.
- `Test (ubuntu-latest, 1.24)`: `watch/TestDebounce` — `Expected 1-2 debounced events, got 4`. That's #135.

Locally on this branch: `TestDebounce` 10/10 pass, `TestRustGraphContextHandlersDisclosePartialCoverage` 5/5 pass, `cmd` 3/3 pass. I can't re-run the jobs myself (`rerun-failed-jobs` returns 403 for me), so I'm merging on the package-isolation evidence rather than a green re-run — flagging that explicitly rather than implying the run went green.

Two leftovers I'm still not blocking on: the redundant early duplicate check at `cmd/context_routing.go:124` (`duplicate && !caseInsensitive` keys on `path` rather than `pathKey`, so it's inert when case-insensitive and identical to the check below it when not), and the case-insensitive `forPrefix` branch materializing all matches before sorting even when `topK` is small. Neither affects correctness.

Merging. Nice result on the benchmark, and thanks for taking the reuse half of the note as seriously as the perf half.

---

Issue #172, opened 2026-09-04
Graph accuracy: fix the four known holes before building on the graph


JordanCoin:
The importers graph is the product (every outside contribution to this repo has been dependency-resolution accuracy). Four open issues show it returning wrong or falsely-complete answers in the ecosystems most users run. This issue tracks them as one unit of work, in order, with one exit test.

## The four holes

| # | Ecosystem | Symptom | Reporter |
|---|---|---|---|
| #132 | TypeScript ESM / NodeNext | `.js` specifier -> `.ts` file edges silently dropped, most imports lost | maintainer |
| #136 | Python | `from .mod import x` resolves to nothing, intra-package edges lost | maintainer |
| #147 | CommonJS | single-quoted `require()` / `import` invisible to `--importers` | external |
| #148 | all | `--deps` / `--importers` / `blast-radius` report `coverage: complete` on languages with no file-level imports | maintainer |

## Order

1. #148 first. Until coverage stops saying `complete` when it is not, every other fix ships behind a status that can lie. This is the "tells you when it doesn't know" promise from the README.
2. #147 (smallest change, external reporter waiting).
3. #136 and #132 (the two largest user populations).

## Exit test for the milestone

- Each issue gets a fixture repo under `testdata/` that fails on `main` today and passes after the fix.
- `codemap --importers` on the fixture returns the exact expected importer list, not a superset or subset.
- `coverage` on a language with no file-level imports is `n/a` or `partial`, never `complete`.
- Release notes state what changed per ecosystem so the ~300 people who download each release can tell if their language is affected.

## Why this comes before anything new

Anything built on the graph (#133 `codemap check`, #134 collision prediction) inherits its honesty and states it with more authority. Fixing the primitives is the prerequisite for both.

Milestone: **Graph accuracy**. Related: #111, #133, #134.

JordanCoin:
## Milestone complete — all five fixes on `main`

Leaving this issue open for you to close; this is the completion record.

| # | Ecosystem | PR | On `main` |
|---|---|---|---|
| #148 | all — falsely `complete` coverage | #174 | `c02e15a` |
| #173 | tsconfig `"@/*": ["./*"]` (create-next-app default) | #175 | `4fc5412` |
| #147 | CommonJS single-quoted `require()` | #176 | `952572a` |
| #136 | Python relative imports | #181 | `ced2e72` |
| #132 | TypeScript ESM `.js` → `.ts` | #182 | `1a7aae7` |

[...]

Issue #173, opened 2026-09-04
scanner: tsconfig alias target "./*" (create-next-app default) never resolves; whole Next.js graph reads as standalone


JordanCoin:
## Symptom

On a stock Next.js project (create-next-app default `tsconfig.json`), `codemap --importers` reports no importers for files that are imported everywhere, and `codemap --deps .` shows nearly every file as standalone. No coverage warning is emitted; the JSON output carries no coverage field at all.

Reproduced 2026-09-04 on JordanCoin/importsentry (Next 16, 86 files) with a `dev` build of `main` at 8cda6ae:

```
$ codemap --importers lib/a1.ts
No files import lib/a1.ts.

$ codemap --importers lib/a1.ts --json
{"root":"...","mode":"importers","file":"lib/a1.ts","importers":null,"importer_count":0,"is_hub":false}
```

Reality (`grep`): 7 files import it via `@/lib/a1`.

## Cause

create-next-app writes:

```json
"paths": { "@/*": ["./*"] }
```

`resolvePathAlias` (scanner/filegraph.go:816) substitutes the wildcard into the target, producing `./lib/a1`, and hands that to `tryExactMatch` / `trySuffixMatch`. The file index has no `./` prefix, so nothing matches. The `./` trim at scanner/filegraph.go:537 only runs on the relative-import path, not on alias targets.

## Proof

Same repo copy, one edit to tsconfig, `"@/*": ["*"]`:

```
$ codemap --importers lib/a1.ts
⚠️  HUB FILE: lib/a1.ts
   Imported by 7 files - changes have wide impact!
```

`--deps .` goes from 28 standalone files to 2.

## Why this matters

`./*` is the default that Next.js has shipped for years, so this is the most common TypeScript layout in the wild, and the failure is silent: the answer looks complete. It compounds #132 (ESM specifiers) and #148 (coverage reports complete). This repo's own README promises tsconfig path aliases are resolved.

## Fix

Normalize alias targets with `path.Clean` (and strip a leading `./`) before matching, and add a fixture under `testdata/` with `"@/*": ["./*"]` that asserts the importer list exactly. Also emit `coverage` in the `--importers` JSON so a zero-importer answer on an aliased project cannot read as complete.

Milestone: Graph accuracy. Tracked in #172. Related: #20 (closed, same symptom, different cause), #132, #148.

PR #174, opened 2026-09-04
fix(coverage): Stop claiming complete coverage on symbol-level languages

### LINKED ISSUE #148: `--deps` / `--importers` / `blast-radius` report `coverage: complete` on languages with no file-level imports

JordanCoin:
First of the **Graph accuracy** milestone (#172). Fixes #148.

## What was wrong

Coverage was derived from one question: *did the scanner run?* ast-grep parses Swift fine — #148 notes 1068 functions extracted correctly — so the source came back `authoritative` and coverage came back `complete`.

But the *edge model* is what doesn't apply. In Swift, Kotlin, Java, C# and Scala an import names a module, package or namespace, never a file, and two files in the same package reference each other with **no import statement at all**. So resolving imports yields `SwiftUI`, `Foundation`, `System.Text` — names belonging to no file in the project — and the graph comes back structurally empty rather than empty-by-accident.

Pairing that with `complete` is what makes it the worst of the bugs: a consumer reads "complete" beside zero dependents and concludes a change is isolated, when nothing capable of finding those edges ever ran.

## The fixture that proves it

`testdata/symbol-imports-swift/` — three Swift files where `UserViewModel` references `User` from `Models.swift`, with no import between them, exactly as Swift requires. On `main` today:

```json
"coverage": {"status": "complete", "sources": [{"name": "ast-grep", "status": "authoritative"}], "issues": []}
```

Zero edges, full confidence. After:

```json
"coverage": {"status": "partial", "sources": [
  {"name": "ast-grep", "status": "authoritative"},
  {"name": "symbol-imports/swift", "status": "unavailable",
   "detail": "Swift (3 files): imports name modules, not files, and same-package files need no import: intra-project edges need symbol-reference resolution and are not represented"}
]}
```

`--importers` inherits it, which also answers the shape of #138 — "No files import X" no longer stands alone when the graph cannot know:

```
No files import Sources/App/Models.swift.
Coverage: partial — Swift: imports name modules, not files, and same-package files need no import: …
```

`testdata/file-imports-go/` is the counterpart guard, so this never becomes a blanket downgrade: Go imports do resolve to files, so an empty graph there is a real finding.

```
--importers pkg/user.go  ->  importers: ["main.go"]  (exact)  coverage: complete
```

`TestSwiftFixtureNeverReportsCompleteCoverage` is load-bearing — against unmodified `filegraph.go` it fails with `coverage sources [{Name:ast-grep Status:authoritative Detail:}] omit a symbol-imports source naming Swift`, which is the bug verbatim.

## What changed

`scanner/importmodel.go` classifies languages by whether an import can resolve to a file. Presence of a symbol-level language downgrades coverage to `partial` and attaches one source per language with its file count — so the output names which slice it cannot see, rather than emitting a bare `partial` with nothing to point at. Wired into the graph's provenance, which is where `--deps`, `--importers` and `blast-radius` all read from, plus the `newDepsProject` default path.

The rule only ever removes confidence: coverage already `partial` or `unavailable` keeps its status, and languages with a file-level import model are untouched.

|---|---|---|
| Swift, Kotlin, Java, C#, Scala | `complete` with 0 edges | `partial`, naming the language and file count |
| Go, Python, JS/TS, Rust, C/C++, Dart, Ruby, PHP, Lua, Solidity, Bash, CUE | unchanged | unchanged |

There is no CHANGELOG in the repo, so this table is the release-note text rather than a file I invented.

## Two things I did not do, deliberately

- **The false edges.** #148's closing note about `VendorSDK ───▶ src/Integrations/VendorSDK` is real: Strategy 6 in `fuzzyResolveWithWorkspace` matches a normalized namespace against a *directory name*, and its comment says it exists for C#. So for these languages codemap can also emit a wrong edge, not just miss them. Removing that is a behavior change for current C# users and deserves its own PR and fixture — say the word and I'll do it next.
- **Elixir** is not classified. `alias MyApp.Foo` is module-level, but Mix conventions do map modules to paths, so suffix matching often works. I'd rather leave it resolvable than wrongly mark a working ecosystem degraded.

## One side effect you should decide on

`testdata/symbol-imports-swift/` contains real `.swift` files, so `codemap .` **on this repo** now reports `partial` with a Swift note. That statement is *true* — the repo does contain 3 unresolvable Swift files — but it's noise from fixtures. The natural fix is `"exclude": ["testdata"]`, except `.codemap/` is gitignored so it can't ship. The alternative is adding `testdata` to `IgnoredDirs` in `scanner/walker.go` alongside `vendor` and `node_modules`, which is a default-behavior change for every user — so I didn't make that call unilaterally. Happy to do either.

## Verification

`go vet ./...` clean. `go test ./...` at the main baseline (only the three known root-environment permission failures). Fixture behavior verified end-to-end with a built binary for `--deps`, `--deps --json` and `--importers`.

---

### DIFF
--- a/blast_radius.go
+++ b/blast_radius.go
@@ -1034,6 +1034,12 @@ func buildBlastRadiusRendered(diffProject scanner.Project, depsProject scanner.D
 	}
 	var renderable []renderableImporter
 	for _, report := range reports {
+		// Decide on the data, not on the rendered text: every report now
+		// carries a coverage line, so rendering first would make a file with
+		// no importers look like it had content.
+		if len(report.Importers) == 0 && len(report.HubImports) == 0 {
+			continue
+		}
 		text := renderImportersReportString(report)
 		if strings.TrimSpace(text) == "" {
 			continue
@@ -1449,7 +1455,7 @@ func buildImportersReportFromGraph(root, file string, fg *scanner.FileGraph) (sc
 		Imports:        imports,
 		ImporterCount:  len(importers),
 		IsHub:          fg.IsHub(file),
-		CoverageStatus: string(fg.Coverage.Status),
+		CoverageStatus: string(fg.Coverage.EffectiveStatus()),
 		CoverageNotes:  append([]string(nil), fg.Coverage.Notes...),
 	}
 
--- a/mcp/analysis_output.go
+++ b/mcp/analysis_output.go
@@ -94,7 +94,7 @@ func newImportersOutput(root, file string, graph *scanner.FileGraph) ImportersOu
 		Kind: kind, Root: root, Mode: "importers", File: file,
 		Importers: boundedImporters, Imports: boundedImports, HubImports: boundedHubs,
 		ImporterCount: len(importers), IsHub: graph.IsHub(file),
-		CoverageStatus: string(graph.Coverage.Status), CoverageNotes: nonNilCopy(graph.Coverage.Notes),
+		CoverageStatus: string(graph.Coverage.EffectiveStatus()), CoverageNotes: nonNilCopy(graph.Coverage.Notes),
 		Truncated: omittedCount > 0, OmittedCount: omittedCount,
 	}
 }
--- a/mcp/structured_contract_test.go
+++ b/mcp/structured_contract_test.go
@@ -228,6 +228,26 @@ func writeParityRepository(t *testing.T) string {
 	return root
 }
 
+// A graph built with nothing to report against it has the zero-value status.
+// The MCP answer must spell that out as "complete": an empty coverage_status
+// makes a zero-importer answer on a fully checked graph indistinguishable from
+// one the scanner could not check at all (#148).
+func TestNewImportersOutputSpellsOutCompleteCoverage(t *testing.T) {
+	graph := &scanner.FileGraph{
+		Importers: map[string][]string{},
+		Imports:   map[string][]string{},
+	}
+	out := newImportersOutput("/repo", "pkg/user.go", graph)
+	if out.CoverageStatus != "complete" {
+		t.Fatalf("coverage_status = %q, want %q for a zero-value graph coverage", out.CoverageStatus, "complete")
+	}
+	graph.Coverage.Status = "partial"
+	out = newImportersOutput("/repo", "pkg/user.go", graph)
+	if out.CoverageStatus != "partial" {
+		t.Fatalf("coverage_status = %q, want %q when the graph reports partial", out.CoverageStatus, "partial")
+	}
+}
+
 func TestNewImportersOutputSortsDeterministically(t *testing.T) {
 	graph := &scanner.FileGraph{
 		Importers: map[string][]string{
--- a/scanner/filegraph.go
+++ b/scanner/filegraph.go
@@ -183,6 +183,11 @@ func buildFileGraphFromAnalysesWithCargoMetadataAndFilters(ctx context.Context, 
 			break
 		}
 	}
+	// Languages whose imports name modules rather than files cannot produce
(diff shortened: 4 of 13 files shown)


JordanCoin:
Pushed `e0c07b7`, folding in the `--importers --json` coverage field — it's asked for by both #148's spirit and #173 explicitly, and it turned out to be a bigger hole than "the field is missing".

`ImportersReport.CoverageStatus` was `omitempty`, and a complete graph's status is the **empty zero value**. So the common case emitted nothing, and on `main` today these two answers are byte-identical:

```json
{"file":"pkg/user.go","importers":null,"importer_count":0,"is_hub":false}   // checked thoroughly, genuinely zero
{"file":"App/Models.swift","importers":null,"importer_count":0,"is_hub":false}   // could not check at all
```

Now:

```json
{"file":"pkg/user.go",...,"coverage_status":"complete"}
{"file":"Sources/App/Models.swift",...,"coverage_status":"partial","coverage_notes":["Swift: imports name modules, not files, …"]}
```

One thing worth calling out, because it's a regression I introduced and then caught rather than something the change did cleanly: making the status always present broke `TestBlastRadiusOmitsEmptyImporterSections` (passes on `main`, failed with my change). Blast-radius decided whether an importer section was empty by *rendering it and checking for whitespace* — and now every report renders a `Coverage: complete` footer, so files with no importers stopped looking empty and started emitting sections. Fixed by deciding on the report's data (`len(Importers) == 0 && len(HubImports) == 0`) instead of its rendered text, which is the sturdier test anyway.

Full suite back at the baseline (only the three root-environment permission failures); `go vet` clean.

---

Issue #147, opened 2026-08-24
js-imports rule only matches double-quoted require()/import — single-quoted CommonJS invisible to --importers


nicknsheth-beep:
## Summary

`scanner/sg-rules/javascript.yml`'s `js-imports` rule only matches **double-quoted** `import`/`require` string literals. Single-quoted CommonJS `require()` calls — extremely common in real-world JS, and the default style under most Prettier/ESLint configs (`singleQuote: true`) — are silently invisible to `--importers` / the MCP `get_importers` tool, which then reports a confident "No files import X" for files that are demonstrably imported elsewhere.

This is a correctness issue, not just a wording one: it produces a *confident false negative* on exactly the kind of blast-radius check someone runs before editing a shared/security-sensitive file.

## Root cause

`scanner/sg-rules/javascript.yml`:
```yaml
id: js-imports
language: javascript
rule:
  any:
    - pattern: import $$$_ from "$PATH"
    - pattern: import "$PATH"
    - pattern: require("$PATH")
```

All three patterns are literal-text ast-grep patterns hardcoded to double-quoted strings. Verified directly with `ast-grep` (0.45.1) against this exact rule:

```js
// single.js
const memberRoutes = require('./routes/members');
```
→ `ast-grep scan --rule rule.yml single.js` produces **zero matches**.

```js
// double.js
const memberRoutes = require("./routes/members");
```
→ produces a match.

By contrast, `scanner/sg-rules/typescript.yml` / `tsx.yml` / `jsx.yml` all use a structural, quote-agnostic rule:
```yaml
id: ts-imports
language: typescript
rule:
  kind: import_statement
```
which correctly matches regardless of quote style. Only the plain `.js` rule uses literal-text patterns, and only for double quotes.

## Impact

On a real Node/Express codebase (100% single-quoted `require()` throughout — the standard style), this means `codemap --importers <any-required-file>.js` reports zero importers for files that have many. Confirmed on two separate repos, e.g. `services/layoutInputService.js` — 14 real requirers via `grep -rn "require(.*layoutInputService"`, `codemap --importers` reports "No files import `services/layoutInputService.js`."

[...]

Issue #132, opened 2026-08-13
scanner: TypeScript ESM/NodeNext projects silently lose most import edges (.js specifier -> .ts file)


JordanCoin:
Found while running an agent fleet against real repositories. On a TypeScript project using ESM / `moduleResolution: NodeNext` — the modern default — codemap loses most of its dependency edges and reports `status: complete` with `issues: []`.

## Reproduction

A correct NodeNext TypeScript project. Under NodeNext the *specifier* must carry a `.js` extension even though the file on disk is `.ts` — this is required by the spec, not a mistake:

```
package.json   {"type": "module"}
tsconfig.json  {"compilerOptions": {"module": "NodeNext", "moduleResolution": "NodeNext"}}

src/types.ts    export type T = { a: number }
src/helper.ts   export function helper() { return 1 }
src/index.ts    import type { T } from "./types.js"
                import { helper } from "./helper.js"
```

```
$ codemap --json --deps .
  index.ts imports: ["./helper.js", "./types.js"]
  coverage: complete | issues: 0

$ codemap --importers src/types.ts
No files import src/types.ts.
   Note: files in the same package never import each other (Go resolves
   imports at package level), so only cross-package importers appear here.
```

`src/types.ts` **is** imported. codemap reports no importers, full confidence, and no issues — then explains itself with a message about Go package semantics on a TypeScript file.

## Scale

Measured on a real 268-file TypeScript repo: codemap reports `93 deps` and `status: complete`. The actual figure is roughly 600 edges — about **515 are lost, ~85%**. Every one of them silently.

This is not an exotic configuration. `"type": "module"` with `moduleResolution: NodeNext` is what `tsc --init` and most current TS tooling produce, and the `.js`-specifier rule is mandatory there.

## Mechanism

`tryExactMatch` (`scanner/filegraph.go:414`) resolves by appending each known extension to the **whole specifier**:

```go
for _, ext := range extensions {
    candidate := path + ext          // "./types.js" + ".ts" = "./types.js.ts"
    if files, ok := idx.byExact[candidate]; ok { ... }
}
```

For `./types.js` it tries `./types.js.ts`, `./types.js.tsx`, … and finally the bare `./types.js` (the empty-string fallback at `scanner/types.go:253`). None exists, so the reference is dropped. The candidate it needs — `./types.ts` — is never constructed, because that requires *replacing* the `.js` suffix rather than appending to it.

[...]

Issue #153, opened 2026-08-24
`blast-radius`'s embedded `[deps]` section is nondeterministic across identical runs


JordanCoin:
**Version:** 4.4.2 (Homebrew) · macOS

## Summary

Running `blast-radius` repeatedly on an unchanged repo at a fixed ref produces different function and dependency counts each time. Only the embedded `[deps]` sub-section varies; the rest of the output is stable.

## Repro

Same repo, same ref, no edits between runs:

```
codemap blast-radius --text --ref <ref> .   # Go repo, 4 consecutive runs
```

## Actual

```
242 functions · 8 deps
174 functions · 3 deps
191 functions · 7 deps
253 functions · 2 deps
```

Same on a Swift repo, and the format also shifts between output modes:

```
--text:      72 functions / 57 / 63
--markdown:  81 / 52 / 81
```

## Scope — this is narrower than it looks, which may help localize it

These are **fully deterministic** (5/5 identical runs each):

- `codemap --deps --json .`
- `codemap --importers <file>` (also verified set-identical to `grep -rl`, 31/31)

And within `blast-radius` itself, the *importer* counters are stable run to run:

```
files_with_dependents=4
impacted_outside_diff=12/24
```

So the standalone deps scanner is fine. Only the copy of it embedded in `blast-radius` races.

## Likely cause

Looks like a concurrent scan that's cut off by `blast-radius`'s character budget or a timeout — whichever workers have finished when the budget is hit is what lands in the output. That would explain why counts vary downward-ish and why the standalone command, which has no such budget, is stable.

## Why it matters

Anything comparing two `blast-radius` runs — CI gating, a diff of before/after, an LLM told "here is the dependency context" — sees changes that didn't happen. Nondeterministic output also makes the tool hard to snapshot-test.

## Suggested fix

Collect results deterministically before applying the budget: gather the full set, sort by a stable key, then truncate. If the truncation is intentional, report it (`showing 174 of N functions`) rather than presenting a partial count as the total.

PR #175, opened 2026-09-04
fix(scanner): Resolve tsconfig alias targets written as "./*"

### LINKED ISSUE #173: scanner: tsconfig alias target "./*" (create-next-app default) never resolves; whole Next.js graph reads as standalone

JordanCoin:
Second of the **Graph accuracy** milestone (#172). Fixes #173.

Your diagnosis was exact, so this mostly confirms it and adds one case you didn't mention.

## What was wrong

`resolvePathAlias` substitutes the wildcard into the target, so `"@/*": ["./*"]` with `@/lib/a1` produces `./lib/a1`. The file index stores repository-relative paths with **no** `./` prefix, so `tryExactMatch` and `trySuffixMatch` both miss and the import resolves to nothing.

`filepath.Join(baseURL, resolved)` cleans the path, which is why this only ever bit projects **without** a `baseUrl` — and why `"@/*": ["*"]` works as the workaround.

The failure is silent. Nothing marks an unresolved alias, so a whole Next.js project reads as standalone files and the answer looks confident.

## The fixture that proves it

`testdata/tsconfig-alias-dotslash/` — create-next-app's default `tsconfig.json`, one `lib/a1.ts`, and two files importing it via `@/lib/a1`. On `main`:

```json
{"file":"lib/a1.ts","importers":null,"importer_count":0,"is_hub":false}
```

After, the importers are `app/layout.tsx` and `app/page.tsx` — both of them, and nothing else. `TestTsconfigDotSlashAliasResolvesImporters` fails against unmodified `resolvePathAlias` with `lib/a1.ts importers = [], want exactly [app/layout.tsx app/page.tsx]`.

**Correction to an earlier version of this description:** it showed that result as an ordered JSON array and called it an "exact list". The membership is exact; the *order* is not stable, and that is a pre-existing property of `main` rather than anything this PR introduces — see the [comment below](https://github.com/JordanCoin/codemap/pull/175#issuecomment-5541407490) for twelve runs producing twelve orderings. The test compares as a set for that reason.

## The case #173 didn't mention

The **no-wildcard exact-alias branch** has the identical bug — `{"@app": ["./lib/a1"]}` never resolved either. Same one-line fix, and it's covered by `TestPathAliasTargetShapes/exact_alias_with_dot_slash`, which also fails on main.

## Every target shape, checked for regression

`TestPathAliasTargetShapes` pins all of them so this fix cannot break the forms that already worked. Against main, exactly the three broken ones fail and the three working ones pass:

| target | main | after |
|---|---|---|
| `"@/*": ["./*"]` | ✗ | ✓ |
| `"@/*": ["./src/*"]` | ✗ | ✓ |
| `"@app": ["./lib/a1"]` | ✗ | ✓ |
| `"@/*": ["*"]` | ✓ | ✓ |
| `"@/*": ["src/*"]` | ✓ | ✓ |
| `"@/*": ["./*"]` + `baseUrl` | ✓ | ✓ |

`normalizeAliasTarget` deliberately preserves `../shared/x` rather than collapsing it, so a target escaping the project can't be rewritten into something that matches an unrelated file.

|---|---|---|
| TypeScript/JavaScript with tsconfig `paths` targets starting `./` (create-next-app default, no `baseUrl`) | aliased imports resolved to nothing; files read as standalone | resolved |
| All other alias shapes, and every other ecosystem | unchanged | unchanged |

## Notes

- The coverage half of #173's fix ("emit `coverage` in the `--importers` JSON") is in **#174**, not here — a zero-importer answer there was byte-identical whether the scan was thorough or impossible. The two fixes are complementary rather than redundant: this PR is a genuine *miss*, so coverage would have said `complete` and been wrong. Disclosure can't substitute for resolution; it only helps where the tool can know it doesn't know.
- **#174 and this PR merge cleanly.** I checked this out and merged both onto `main` locally: clean merge, `go build` and `go vet` clean, suite at baseline.
- The importer-ordering nondeterminism this surfaced is **not fixed here** — sorting `Imports`/`Importers` changes user-visible output for everyone and wants its own PR and fixture. It does block #172's "exact importer list" exit criterion, so it's worth scheduling.
- I did not touch #132 (ESM `.js` → `.ts`), which #173 notes it compounds. That's still queued behind #147.

## Verification

`go vet ./...` clean, `gofmt` clean, `go test ./...` at the main baseline (only the three known root-environment permission failures). Fixture verified end-to-end with a built binary for `--importers` and `--importers --json`, and the fixture test passes at `-count=20`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1153,22 +1153,31 @@ func publishWatchReadiness(path string, readinessErr error) error {
 
 func waitWatchReadiness(path string, timeout time.Duration) error {
 	deadline := time.Now().Add(timeout)
+	// A readiness file that does not parse is a file still being written, not
+	// a daemon that failed: publishWatchReadiness renames its payload into
+	// place atomically, but nothing guarantees every writer does, and treating
+	// the first unparseable read as fatal reported a startup failure for a
+	// daemon that was starting fine. Keep the last parse error so a file that
+	// never becomes valid says why, rather than only that it timed out.
+	var lastParseErr error
 	for {
 		data, err := os.ReadFile(path)
 		if err == nil {
 			var status watchReadiness
-			if err := json.Unmarshal(data, &status); err != nil {
-				return fmt.Errorf("reading daemon readiness: %w", err)
-			}
-			if status.Error != "" {
+			if parseErr := json.Unmarshal(data, &status); parseErr != nil {
+				lastParseErr = parseErr
+			} else if status.Error != "" {
 				return errors.New(status.Error)
+			} else {
+				return nil
 			}
-			return nil
-		}
-		if !errors.Is(err, os.ErrNotExist) {
+		} else if !errors.Is(err, os.ErrNotExist) {
 			return fmt.Errorf("reading daemon readiness: %w", err)
 		}
 		if time.Now().After(deadline) {
+			if lastParseErr != nil {
+				return fmt.Errorf("reading daemon readiness: %w", lastParseErr)
+			}
 			return fmt.Errorf("daemon readiness timed out after %s", timeout)
 		}
 		time.Sleep(10 * time.Millisecond)
--- a/scanner/aliastarget_test.go
+++ b/scanner/aliastarget_test.go
@@ -3,6 +3,7 @@ package scanner
 import (
 	"context"
 	"reflect"
+	"sort"
 	"testing"
 )
 
@@ -15,7 +16,11 @@ func TestTsconfigDotSlashAliasResolvesImporters(t *testing.T) {
 	if err != nil {
 		t.Fatalf("build alias fixture graph: %v", err)
 	}
-	got := graph.Importers["lib/a1.ts"]
+	// Compare as a set: the graph appends importers in analysis order, which
+	// is not deterministic across runs (see the note in the PR for #173). The
+	// exactness this asserts is membership, not sequence.
+	got := append([]string(nil), graph.Importers["lib/a1.ts"]...)
+	sort.Strings(got)
 	want := []string{"app/layout.tsx", "app/page.tsx"}
 	if !reflect.DeepEqual(got, want) {
 		t.Fatalf("lib/a1.ts importers = %v, want exactly %v", got, want)
--- a/scanner/aliastarget_test.go
+++ b/scanner/aliastarget_test.go
@@ -1,0 +1,78 @@
+package scanner
+
+import (
+	"context"
+	"reflect"
+	"testing"
+)
+
+// create-next-app has shipped "@/*": ["./*"] for years, so this is the most
+// common TypeScript layout in the wild. The substituted target was "./lib/a1"
+// while the file index holds "lib/a1", so nothing matched and a file imported
+// everywhere reported no importers at all.
+func TestTsconfigDotSlashAliasResolvesImporters(t *testing.T) {
+	graph, err := BuildFileGraph(context.Background(), "../testdata/tsconfig-alias-dotslash", Filters{})
+	if err != nil {
+		t.Fatalf("build alias fixture graph: %v", err)
+	}
+	got := graph.Importers["lib/a1.ts"]
+	want := []string{"app/layout.tsx", "app/page.tsx"}
+	if !reflect.DeepEqual(got, want) {
+		t.Fatalf("lib/a1.ts importers = %v, want exactly %v", got, want)
+	}
+}
+
+func TestNormalizeAliasTarget(t *testing.T) {
+	for _, tc := range []struct {
+		name   string
+		target string
+		want   string
+	}{
+		{"create-next-app default", "./lib/a1", "lib/a1"},
+		{"bare wildcard target", "lib/a1", "lib/a1"},
+		{"nested with dot slash", "./src/lib/s1", "src/lib/s1"},
+		{"redundant separators", "./src//lib/../lib/s1", "src/lib/s1"},
+		{"bare dot collapses to root", ".", ""},
+		{"dot slash only", "./", ""},
+		{"empty stays empty", "", ""},
+		// A target escaping the project must not be silently rewritten into a
+		// path that could match an unrelated file.
+		{"parent traversal preserved", "../shared/x", "../shared/x"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			if got := normalizeAliasTarget(tc.target); got != tc.want {
+				t.Fatalf("normalizeAliasTarget(%q) = %q, want %q", tc.target, got, tc.want)
+			}
+		})
+	}
+}
+
+// Every alias target shape has to keep resolving, so the "./" fix cannot
+// regress the forms that already worked.
+func TestPathAliasTargetShapes(t *testing.T) {
+	files := []FileInfo{{Path: "lib/a1.ts"}, {Path: "src/lib/s1.ts"}, {Path: "app/page.tsx"}}
+	idx := buildFileIndex(files, "")
+
+	for _, tc := range []struct {
+		name    string
+		imp     string
+		aliases map[string][]string
+		baseURL string
+		want    []string
+	}{
+		{"dot slash wildcard", "@/lib/a1", map[string][]string{"@/*": {"./*"}}, "", []string{"lib/a1.ts"}},
+		{"bare wildcard", "@/lib/a1", map[string][]string{"@/*": {"*"}}, "", []string{"lib/a1.ts"}},
+		{"nested dot slash", "@/lib/s1", map[string][]string{"@/*": {"./src/*"}}, "", []string{"src/lib/s1.ts"}},
+		{"nested bare", "@/lib/s1", map[string][]string{"@/*": {"src/*"}}, "", []string{"src/lib/s1.ts"}},
+		{"base url with dot slash", "@/lib/a1", map[string][]string{"@/*": {"./*"}}, ".", []string{"lib/a1.ts"}},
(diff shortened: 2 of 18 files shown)


JordanCoin:
CI went red on `Test (macos-latest, 1.24)` and `Test (ubuntu-latest, 1.26)` — my test, not the fix. Pushed `298dffd`. But what it caught is worth more than the fix itself.

```
aliastarget_test.go:21: lib/a1.ts importers = [app/page.tsx app/layout.tsx],
                        want exactly [app/layout.tsx app/page.tsx]
```

Same members, different order. I'd asserted a sequence, so the test was reading a real property of the output as a failure.

## `--importers` output order is nondeterministic on `main`

Not caused by this PR. Twelve consecutive runs of the same binary against the same unmodified fixture, on `8cda6ae`:

```
app/sidebar.tsx,app/header.tsx,app/layout.tsx,app/footer.tsx,app/page.tsx
app/page.tsx,app/header.tsx,app/sidebar.tsx,app/footer.tsx,app/layout.tsx
app/page.tsx,app/header.tsx,app/layout.tsx,app/footer.tsx,app/sidebar.tsx
app/page.tsx,app/footer.tsx,app/header.tsx,app/layout.tsx,app/sidebar.tsx
app/layout.tsx,app/sidebar.tsx,app/page.tsx,app/header.tsx,app/footer.tsx
app/layout.tsx,app/sidebar.tsx,app/footer.tsx,app/page.tsx,app/header.tsx
app/header.tsx,app/footer.tsx,app/page.tsx,app/sidebar.tsx,app/layout.tsx
app/header.tsx,app/footer.tsx,app/page.tsx,app/layout.tsx,app/sidebar.tsx
app/header.tsx,app/footer.tsx,app/layout.tsx,app/page.tsx,app/sidebar.tsx
app/footer.tsx,app/layout.tsx,app/sidebar.tsx,app/header.tsx,app/page.tsx
app/footer.tsx,app/header.tsx,app/page.tsx,app/sidebar.tsx,app/layout.tsx
app/footer.tsx,app/header.tsx,app/layout.tsx,app/sidebar.tsx,app/page.tsx
```

Twelve runs, twelve orderings. `fg.Importers[imported] = append(...)` follows analysis order, and `scanner/filegraph.go` sorts nothing — there is no `sort` in the file at all. `Imports` has the same shape.

This matters for #172 beyond cosmetics: the milestone's exit test is *"`codemap --importers` on the fixture returns the exact expected importer list"*, and no exact list assertion can hold while the sequence reshuffles every run. It's also the same family as #153 (`blast-radius`'s `[deps]` section nondeterministic across identical runs), and it means anyone diffing codemap output between runs — or feeding it to an LLM expecting stable context — sees churn that isn't real.

## What I did and didn't do

Here I only made the test compare **membership** rather than sequence, which is the exactness that this PR's fix is actually about (not a superset, not a subset). Sorting `Imports`/`Importers` in `FileGraph` is a two-line change and I'd expect it to be the right one, but it changes user-visible output ordering for everyone, so it wants its own PR, its own fixture, and your call — not a quiet ride-along in an alias fix. Say the word and it's next, before or after #147.

Full suite back at baseline; `go vet` and `gofmt` clean; the fixture test passes at `-count=20`.

---

JordanCoin:
`Test (ubuntu-latest, 1.24)` is red on `298dffd`, and it's a different failure from the ordering one — `codemap/scanner` passed, so the alias fix and its test are fine. **This one is not this PR's, and it's a real bug rather than a flaky test.**

```
main_more_test.go:311: runWatchSubcommand(start) error =
  starting daemon: reading daemon readiness: unexpected end of JSON input,
  want child readiness failure
```

## Not this PR's

This diff touches `scanner/filegraph.go` and a fixture. The failure is in the root package's watch-start plumbing. Compiled test binaries from both branches, alternated under CPU contention, 20 runs each:

```
main   : 20/20 failures
pr175  : 20/20 failures
```

Identical. Unloaded it's 0/15 on both, which is why local runs and most CI runs are green.

## What's actually broken

`waitWatchReadiness` (`main.go:1154`) polls the readiness file and treats *any* successful read as final:

```go
data, err := os.ReadFile(path)
if err == nil {
    var status watchReadiness
    if err := json.Unmarshal(data, &status); err != nil {
        return fmt.Errorf("reading daemon readiness: %w", err)   // ← gives up here
    }
```

A file that exists but is empty or half-written reads successfully with zero or partial bytes, so `json.Unmarshal` fails and the wait aborts. Only `os.ErrNotExist` is treated as "not ready yet".

Proved directly against the real function:

```
empty readiness file   -> err=reading daemon readiness: unexpected end of JSON input after 0s
partial readiness file -> err=reading daemon readiness: unexpected end of JSON input
```

Note the **0s** — it doesn't wait out any part of the 30s timeout, it fails on the first poll. In the test the child is `sh -c 'printf ... > $CODEMAP_WATCH_READINESS_FILE'`, and `>` truncates the file into existence before `printf` writes it; under load the poller lands in that window.

This isn't confined to the test. Any daemon that creates its readiness file non-atomically has the same window, so on a loaded machine `codemap watch start` can fail with `unexpected end of JSON input` for a daemon that started perfectly well — reporting a startup failure that didn't happen.

## What I'm doing about it

Not fixing it inside this PR — it's `main.go` watch plumbing with no relationship to tsconfig aliases, and widening an alias fix to carry it would be the wrong shape. I'm opening it as its own small PR instead, so it fixes the root cause for every PR rather than just unblocking this one. Two halves:

- **Reader:** treat an unparseable readiness file as *not ready yet* and keep polling until the deadline, rather than failing on the first poll. That's the robust half — it also covers a daemon writing non-atomically.
- **Writer:** write the readiness file atomically via the existing `runtimefile.WriteAtomic`, so the window doesn't exist in the first place.

`rerun-failed-jobs` returns 403 for me, so I can't re-run this leg. I'll link the fix PR here when it's up and keep this one watched until it's green.

---

JordanCoin:
Correcting two things in my comment above before they mislead anyone.

**1. The "20/20 failures on both branches" measurement was wrong.** I ran relocated test binaries from outside the repo; from there they exit without running the test at all, so I was measuring my own harness, not the test. Redone properly — in-repo, via `go test`, under CPU contention — it's **0/10 on both `main` and this branch**, and 15/15 pass unloaded. **I cannot reproduce the CI failure locally at all.**

The conclusion that it isn't this PR's still holds, on better evidence: the diff touches only `scanner/filegraph.go` and a fixture, `codemap/scanner` passed in that same CI run, and the failing test is in the root package's watch-start plumbing.

**2. I overstated the production impact.** I wrote that "on a loaded machine `codemap watch start` can fail for a daemon that started perfectly well". `publishWatchReadiness` already writes via temp-file + `os.Rename`, so codemap's own daemon never opens that window. The brittleness is reader-side only — real, but defence in depth rather than a user-facing startup bug.

What survives both corrections: the mechanism is confirmed. An empty or partial readiness file makes `waitWatchReadiness` return `reading daemon readiness: unexpected end of JSON input` after **0s**, and that's the only code path producing the string CI reported. The test's child is `sh -c 'printf … > "$CODEMAP_WATCH_READINESS_FILE"'`, and `>` truncates the file into existence before `printf` writes — exactly that window.

Fix is up as #177, standalone. It's right on its own terms — an unparseable file is a file still being written, not a failed daemon — but I'm not claiming it as proven to fix this leg, since I can't reproduce the failure to test against.

This PR's own head `298dffd` has `codemap/scanner` green and the alias fix intact. `rerun-failed-jobs` still 403s for me, so I can't re-run the leg; keeping it watched.

---

PR #179, opened 2026-09-04
feat(scanner): Skip testdata directories by default

JordanCoin:
Adds `testdata` to the hardcoded ignore list, alongside `vendor` and `node_modules`. Relates to #172.

## Why

Fixture repositories under `testdata` are not the project's source. They are checked-in inputs for tests: scanning them counts fixture files as project files, and a fixture written in a language the file-level edge model does not cover drags the *project's* reported coverage down with it.

Go's own toolchain takes the same position — `go build` and the `go` tool ignore `testdata` entirely, and have since the beginning — so this matches a convention Go users already have. It is the justification for the default; the coverage effect is the consequence.

An earlier draft of this PR showed a coverage `partial` → `complete` before/after. That measurement came from a Swift fixture added by #174, which is not merged, so it does not reproduce on `main` — it has been removed rather than left as an unverifiable claim.

## What actually changes on this repo

`main`'s tracked tree contains exactly one `testdata` directory: `watch/testdata/`, seven shell scripts (`run_test.sh`, `test_mcp.sh`, and five under `output/`). Measured with binaries built from `origin/main` and from this branch, both scanning a clean export of `origin/main`:

```
main branch:  Files: 283   structure output contains  watch/testdata/  (7 files, 17.4KB, all .sh)
this branch:  Files: 276   no testdata node in the tree
```

So the visible effect today is modest and mostly forward-looking: seven shell fixtures leave the file count, and the bare `watch/testdata/` directory node leaves the tree. The value is in what it prevents as fixture directories accumulate — this repo's own milestone work adds them steadily.

## Can it be turned off? No — and that is worth being explicit about

`IgnoredDirs` is a hardcoded map in `scanner/walker.go`, checked on the fast path *before* gitignore and before any project config. `Only` and `Exclude` in `.codemap/config.json` are additive filters — they narrow what a scan returns, they cannot widen it — so there is no config key, flag, or gitignore negation that puts `testdata` back into a scan.

The one escape hatch is scanning the fixture directly as the root, and this PR now guarantees that: `IgnoredDirs` is no longer applied to the walk root, so `codemap testdata/`, or `codemap .` from inside `vendor/`, scans the directory the user actually named. Before that fix the walker matched the root's own base name and returned `Files: 0` — no error, no message, just an empty result. `ReadExternalDeps` gets the same exemption so a manifest sitting at that root is still read.

Anyone who deliberately keeps project source under a directory named `testdata` will see those files disappear from counts and from the tree, with no way to opt back in short of renaming the directory or scanning it directly. Flagging it plainly: it is a default-behaviour change with no override.

## Tests

- `TestScanSkipsTestdataDirectories` — skipped at top level and nested (`internal/testdata/nested/`).
- `TestTestdataFixtureScansWhenItIsTheRoot` — a fixture passed as the scan root still scans, which is how the fixture tests in `scanner` call `BuildFileGraph`.
- `TestIgnoredDirNameScansWhenItIsTheRoot` — a root literally named `testdata`, `vendor`, or `node_modules` scans its files. Fails on the pre-fix walker with `returned [], want [app.go]` for all three.
- `TestIgnoredDirRootStillSkipsNestedIgnoredDirs` — the root exemption covers the root only; a nested `testdata` under a `testdata` root is still skipped.

## Also on this branch

Two fixes ported from open PRs so this branch's CI can go green without waiting on them; both no-op once `main` carries them.

- Cargo metadata timeout (`scanner/rustcargo.go`), from @reneleonhardt's #171: `buildRustWorkspaceIndex` shadowed its caller's `ctx` with the cargo-metadata deadline, so an expired deadline failed the whole graph build instead of falling back to the manually derived Rust workspace.
- Watch readiness (`main.go`), the window #177 fixes: `waitWatchReadiness` gave up on the first unparseable read of a readiness file still being written, surfacing as `reading daemon readiness: unexpected end of JSON input`.

## Release note

`testdata` directories are now skipped by default, matching Go's own toolchain convention and joining `vendor` and `node_modules`. Fixture files no longer count as project files or affect a project's reported coverage. Passing such a directory as the scan root still scans it — and now does so for every name on the ignore list, where previously it returned no files.

## Verification

`go vet ./...` clean, `gofmt` clean, `go test ./...` fully green on this branch.

https://claude.ai/code/session_01TcyheQmM3HCvxF5wRL3s5t

### DIFF
--- a/blast_radius.go
+++ b/blast_radius.go
@@ -1034,6 +1034,12 @@ func buildBlastRadiusRendered(diffProject scanner.Project, depsProject scanner.D
 	}
 	var renderable []renderableImporter
 	for _, report := range reports {
+		// Decide on the data, not on the rendered text: every report now
+		// carries a coverage line, so rendering first would make a file with
+		// no importers look like it had content.
+		if len(report.Importers) == 0 && len(report.HubImports) == 0 {
+			continue
+		}
 		text := renderImportersReportString(report)
 		if strings.TrimSpace(text) == "" {
 			continue
@@ -1449,7 +1455,7 @@ func buildImportersReportFromGraph(root, file string, fg *scanner.FileGraph) (sc
 		Imports:        imports,
 		ImporterCount:  len(importers),
 		IsHub:          fg.IsHub(file),
-		CoverageStatus: string(fg.Coverage.Status),
+		CoverageStatus: string(fg.Coverage.EffectiveStatus()),
 		CoverageNotes:  append([]string(nil), fg.Coverage.Notes...),
 	}
 
--- a/main.go
+++ b/main.go
@@ -1153,22 +1153,31 @@ func publishWatchReadiness(path string, readinessErr error) error {
 
 func waitWatchReadiness(path string, timeout time.Duration) error {
 	deadline := time.Now().Add(timeout)
+	// A readiness file that does not parse is a file still being written, not
+	// a daemon that failed: publishWatchReadiness renames its payload into
+	// place atomically, but nothing guarantees every writer does, and treating
+	// the first unparseable read as fatal reported a startup failure for a
+	// daemon that was starting fine. Keep the last parse error so a file that
+	// never becomes valid says why, rather than only that it timed out.
+	var lastParseErr error
 	for {
 		data, err := os.ReadFile(path)
 		if err == nil {
 			var status watchReadiness
-			if err := json.Unmarshal(data, &status); err != nil {
-				return fmt.Errorf("reading daemon readiness: %w", err)
-			}
-			if status.Error != "" {
+			if parseErr := json.Unmarshal(data, &status); parseErr != nil {
+				lastParseErr = parseErr
+			} else if status.Error != "" {
 				return errors.New(status.Error)
+			} else {
+				return nil
 			}
-			return nil
-		}
-		if !errors.Is(err, os.ErrNotExist) {
+		} else if !errors.Is(err, os.ErrNotExist) {
 			return fmt.Errorf("reading daemon readiness: %w", err)
 		}
 		if time.Now().After(deadline) {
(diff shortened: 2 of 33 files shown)


JordanCoin:
The review findings on this PR's description are correct and both errors are mine. Not pushing while the branch is being patched — recording the acknowledgement.

**The before/after in my description cannot be reproduced on `main`.** I wrote:

```
before:  codemap --deps .  ->  coverage: partial, source: symbol-imports/swift
after:   codemap --deps .  ->  coverage: complete
```

That `partial` depends on **#174**, which is unmerged — it's #174's Swift fixture and #174's symbol-level coverage rule that produce it. Anyone checking against `main` would find the premise absent. The effect this PR actually has on `main` today is on the file inventory: `testdata` paths in the tree drop from 11 to 1. I built that framing while working with #174's changes in my tree and never re-derived it from a clean `main`, which is precisely the trap of describing behaviour from a working copy rather than from the base.

**My "scan a fixture directly by passing it as the root" claim was under-tested.** I verified `testdata/probe` as root, where the root's basename is `probe` — so `IgnoredDirs` never sees it. I did **not** try `testdata/` *itself* as root, where the basename is the ignored name and the scan silently returns zero files. `testdata` is a plausible working directory in a way `vendor` isn't, so the pre-existing behaviour matters more here. Making the walker not apply `IgnoredDirs` to the root directory itself is the right fix, and it should carry a test — my `TestTestdataFixtureScansWhenItIsTheRoot` passes without covering the case that actually breaks.

The underlying change stands: fixture repositories aren't project source, and Go's toolchain ignores the directory for the same reason. It's the description and the test's coverage that were wrong, not the exclusion.

I'll re-verify the restated body and the root-directory fix against a clean `main` build once the branch is free, rather than reading the diff and assuming.

---

